PR #5, opened 2024-07-02
resolve dependency-file-generator warning

jameslamb:
## Description

Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Uses `--file-key` instead of `--file_key` in `rapids-dependency-file-generator` calls in CI scripts, as `file_key` is deprecated and set to be removed.


alexbarghi-nv (approved):
👍

PR #7, opened 2024-07-03
Avoid --find-links in wheel jobs

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/69
Contributes to https://github.com/rapidsai/build-planning/issues/33

https://github.com/rapidsai/cugraph/pull/4509 proposed a stricter pattern for passing a `pylibcugraph` wheel from the `wheel-build-cpp` job that produced it into the `wheel-build-python` job wanting to use it (as a build dependency of `cugraph`). This change improves the likelihood that issues with the `pylibcugraph` wheels will be caught in CI.

This proposes porting that change here, so that when CI is set up in this repo, it'll also benefit from that strictness.

### DIFF
--- a/ci/build_wheel_cugraph.sh
+++ b/ci/build_wheel_cugraph.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


jameslamb:
> I'm not sure we need this?

By "this" you mean the use of `rapids-download-wheels-from-s3` to download a wheel? That makes sense, I suspected that maybe that was just copied over from `cugraph` and not yet changed here.

You can close this if you'd like.

But if you do plan to keep using the "download wheels from S3 approach", then something like what I've put up here will more reliably catch issues in those wheels than the `--find-links` approach.

alexbarghi-nv:
I think we should remove that file since we aren't building wheels for cugraph here.  Thanks James.

alexbarghi-nv:
I'll make a PR removing the unneeded script.

jameslamb:
Ok thanks, no prob! We can close this then.

PR #34, opened 2024-08-28
upgrade target Python version for black to 3.10

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/88.

Fixes one Python 3.9 to 3.10 upgrade that was missed in #32.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: black
         language_version: python3
-        args: [--target-version=py39]
+        args: [--target-version=py310]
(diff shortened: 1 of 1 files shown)


jakirkham (approved):
Thanks James! 🙏

Let's raise an issue on benchmarks & notebooks needing similar changes

jameslamb:
> Let's raise an issue on benchmarks & notebooks needing similar changes

Do you see any specific examples in this repo that need that?

I only see CHANGELOG entries (which should not be changed):

```shell
git grep -E '3\.8|3\.9|py38|py39'
```

I did open https://github.com/rapidsai/cugraph/issues/4640 over in `cugraph` for some examples there that could be updated.

PR #43, opened 2024-09-20
Fix update-version.sh

AyodeAwe:


ci/release/update-version.sh
```diff
@@ -77,36 +77,12 @@ for DEP in "${DEPENDENCIES[@]}"; do
...
-    sed_runner "s@rapidsai/devcontainers/features/ucx:[0-9.]*@rapidsai/devcontainers/features/ucx:${NEXT_SHORT_TAG_PEP440}@" "${filename}"
-    sed_runner "s@rapidsai/devcontainers/features/cuda:[0-9.]*@rapidsai/devcontainers/features/cuda:${NEXT_SHORT_TAG_PEP440}@" "${filename}"
-    sed_runner "s@rapidsai/devcontainers/features/rapids-build-utils:[0-9.]*@rapidsai/devcontainers/features/rapids-build-utils:${NEXT_SHORT_TAG_PEP440}@" "${filename}"
-    sed_runner "s@rapids-\${localWorkspaceFolderBasename}-[0-9.]*@rapids-\${localWorkspaceFolderBasename}-${NEXT_SHORT_TAG}@g" "${filename}"
-done
-
-sed_runner "s/:[0-9][0-9]\.[0-9][0-9]/:${NEXT_SHORT_TAG}/" ./notebooks/README.md
-sed_runner "s/branch-[0-9][0-9].[0-9][0-9]/branch-${NEXT_SHORT_TAG}/" ./docs/cugraph/source/nx_cugraph/nx_cugraph.md
+sed_runner "s/\(PROJECT_NUMBER[[:space:]]*\)=.*/\1= ${NEXT_SHORT_TAG}/" cpp/Doxyfile
+sed_runner "s/set(RAPIDS_VERSION *\"[0-9.]*\")/set(RAPIDS_VERSION \"${NEXT_SHORT_TAG}\")/" cpp/CMakeLists.txt
```
jameslamb, line 87: Instead of this, I think we should rive the rapids version from the `VERSION` file to reduce duplication:

* https://github.com/rapidsai/cugraph/pull/4217
* https://github.com/rapidsai/build-planning/issues/15

jameslamb (approved):
Approving this is an improvement on the current state. Left one suggestion for making this even simpler and reducing more sources of possible inconsistencies... but that can be done in a later PR if you'd prefer.

AyodeAwe:
Yep, will fix this in a separate PR.

PR #51, opened 2024-10-10
[NO MRG] Test CI

jakirkham:


jameslamb:
This was opened last week while CI was initially being set up in this repo. We now have that (#56, for example) and are working on the rest in #53 and associated PRs. Closing this.

PR #53, opened 2024-10-11
Add CI for cuGraph-GNN

alexbarghi-nv:
Adds CI for cuGraph-GNN.

Includes changes from these:

* #52 
* https://github.com/rapidsai/wholegraph/pull/228
* https://github.com/rapidsai/cugraph/pull/4701
* https://github.com/rapidsai/cugraph/pull/4661
* https://github.com/rapidsai/cugraph/pull/4615
* https://github.com/rapidsai/wholegraph/pull/215
* https://github.com/rapidsai/cugraph/pull/4393

PR #56, opened 2024-10-15
introduce minimal CI for PRs

jameslamb:
Proposing breaking some changes off of #53, to reduce the size of that PR and get some CI running on PRs in this repo.


ci/build_python.sh
```diff
@@ -42,16 +41,6 @@ rapids-conda-retry mambabuild \
   --channel "${RAPIDS_CONDA_BLD_OUTPUT_DIR}" \
   conda/recipes/cugraph
```
bdice, line 42: Why does this repo build cugraph? Shouldn’t it only build GNN related packages?

dependencies.yaml
```diff
@@ -3,7 +3,7 @@ files:
   all:
     output: [conda]
     matrix:
-      cuda: ["11.8", "12.1"]
+      cuda: ["11.8", "12.1", "12.4"]
```
bdice, line 6: Is this set of versions chosen to align with PyTorch instead of RAPIDS releases? Can we add a comment about that?

```suggestion
      # CUDA versions match PyTorch releases, not RAPIDS
      cuda: ["11.8", "12.1", "12.4"]
```

ci/build_python.sh
jameslamb: You're right, this repo should only be building `cugraph-dgl`, `cugraph-pyg`, `libwholegraph`, and `pylibwholegraph`.

Since you commented on it, I just pushed that change to this specific script.

If you look around this repo you'll see many other unnecessary things. This repo was created by copying big chunks of the `cugraph` and `wholegraph` repos, and there's lots of such cleanup needed. #53 is trying to do that all at once and I think it's dragging as a result... this PR is just trying to take a small, incremental step forward to reduce the size of that effort.

dependencies.yaml
jameslamb: I don't know, just copied this from #53. @alexbarghi-nv can you answer this?

ci/build_python.sh
bdice: Oh. I realize now that this PR isn't intending to do builds at all, it is really "minimal" CI (just style checks). In that case, I'll approve this as-is.

bdice (approved):
Approving this so we can get style checks going (very minimal CI).

dependencies.yaml
jameslamb: I'm gonna merge PR this so we can keep making progress towards setting CI up for this repo, but I agree with you that this needs a comment. Will update it once @alexbarghi-nv has a chance to respond here.

alexbarghi-nv: Sorry, yeah, this is meant to align with PyTorch versions

alexbarghi-nv: Sorry, yeah, this is meant to align with PyTorch versions

jameslamb: no prob, thanks for that! I just added this comment over in the next PR: #57

PR #60, opened 2024-10-21
remove docs support in build.sh, other small cleanup

jameslamb:
Pulling more changes off of #53, to make that diff smaller and get us closer to having full CI running in this repo.

* removes `docs` argument for `build.sh` (docs building is handled by `ci/build_docs.sh`)
* removes unnecessary `wheel_smoke_test_cugraph.py` (copied over from `cugraph` but not needed here, to be replaced by other scripts you can see in the diff of #53)
* other minor cleanup of CI scripts


ci/test_wheel.sh
```diff
@@ -1,40 +0,0 @@
-#!/bin/bash
```
jameslamb, line 1: @alexbarghi-nv this looks like it's not used on #53.

I strongly suspect the version here was adopted from https://github.com/rapidsai/cugraph/blob/branch-24.12/ci/test_wheel.sh, which in the `cugraph` repo is never used with `cugraph-dgl` / `cugraph-pyg`: https://github.com/search?q=repo%3Arapidsai%2Fcugraph%20%22test_wheel.sh%22&type=code

I strongly suspect `test_wheel.sh` could just be removed, and that the `ci/wheel_smoke_test_${package_name}.py` scripts being proposed on #53 are unnecessary, and covering functionality already covered by the `ci/test_wheel_${package_name}.sh` scripts.

alexbarghi-nv: Yeah, this can be taken out

jameslamb: Thanks! I've just pushed changes to #53 removing the smoke-testing stuff that was added there.

PR #61, opened 2024-10-30
add nightly builds/tests, remove some cugraph references

jameslamb:
Continues the work of setting up CI here (breaking more changes off of #53).

Proposes adding nightly builds / tests, but **not publishing** to the nightly PyPI index / `rapidsai-nightly` conda channel. Those publishing jobs should be added here only once `cugraph` folks are ready to fully move development of these libraries to this repo. ref: https://github.com/rapidsai/cugraph-gnn/pull/58#discussion_r1806797656

This also removes some minor lingering cugraph references that look like they don't belong in this repo.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,75 @@
...
+    needs: [cpp-build]
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.12
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      sha: ${{ inputs.sha }}
+  wheel-build-cugraph-dgl:
+    needs: wheel-publish-pylibwholegraph
```
bdice, line 48: The `wheel-publish-pylibwholegraph` job isn't defined here. This probably needs to be removed.
```suggestion
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,75 @@
...
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.12
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      sha: ${{ inputs.sha }}
+      date: ${{ inputs.date }}
+      script: ci/build_wheel_cugraph-dgl.sh
+  wheel-build-cugraph-pyg:
+    needs: wheel-publish-pylibwholegraph
```
bdice, line 58: The `wheel-publish-pylibwholegraph` job isn't defined here. This probably needs to be removed.
```suggestion
```

.github/workflows/build.yaml
jameslamb: blegh missed these, thank you

.github/workflows/build.yaml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  cpp-build:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-24.12
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      sha: ${{ inputs.sha }}
+      node_type: cpu32
```
bdice, line 37: Do you think we still need `cpu32` here? Default is `cpu16`, I think. We use `cpu32` for cugraph because it's quite large with sccache misses. Are the GNN packages equally large? I would guess they are faster to compile than cugraph but don't know for sure.
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -0,0 +1,62 @@
...
+  wheel-tests-pylibwholegraph:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-24.12
+    with:
+      build_type: nightly
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      sha: ${{ inputs.sha }}
+      script: ci/test_wheel_pylibwholegraph.sh
+      matrix_filter: map(select(.ARCH == "amd64"))
```
bdice, line 42: Just checking - do we know why all these packages skip ARM tests? Should we comment about that?

bdice (approved):

.github/workflows/build.yaml
jameslamb: That's an excellent point! I think you're right that this doesn't need `cpu32`. `conda-build-cpp` is just building `libwholegraph`, and it accepts the default runner when built over in the `wholegraph` repo: https://github.com/rapidsai/wholegraph/blob/6821f7ad99b4418ea59887249720469fa676bad2/.github/workflows/build.yaml#L29-L37

.github/workflows/test.yaml
jameslamb: I have no idea.

@BradReesWork added that for `wholegraph` over a year ago: https://github.com/rapidsai/wholegraph/pull/24

@tingyu66 added that constraint for `cugraph-dgl` and `cugraph-pyg` about 9 months ago: https://github.com/rapidsai/cugraph/pull/4133

Maybe they or @alexbarghi-nv could share why it's like that.

.github/workflows/build.yaml
jameslamb: Just pushed https://github.com/rapidsai/cugraph-gnn/pull/61/commits/805062d33f7013865870992ec3956c548a847dff changing these builds and those on PRs to use whatever the default runner size is for the build workflows.

.github/workflows/test.yaml
alexbarghi-nv: There are no packages available on arm64 for pytorch.

alexbarghi-nv: I tried re-enabling those tests and they all failed again because it pulled the pytorch CPU package.

alexbarghi-nv: As far as I know, most ARM64 users of pytorch are going through DLFW or some other means.

jameslamb: alright thank you. I do see `aarch64` CUDA builds of `pytorch` 2.4.1 (the latest version) at https://anaconda.org/conda-forge/pytorch/files, so maybe in the future there's an opportunity to turn these tests back on.

I'll leave it as-is for the purpose of this PR yet, to get more CI up and running so we can move development over here from the `cugraph` repo.

PR #73, opened 2024-11-19
[Bugfix] Add stream synchronization before the scatter operation

chang-l:
This is to address the issue from this PR: https://github.com/rapidsai/wholegraph/pull/229,  and it's only for the last scatter operation before the Python interface (not for all internal `scatter_func` calls)

Since the output of the scatter operation could be on the host (e.g., when emb_device = 'cpu'), it is necessary to perform synchronization internally. This ensures users do not need to explicitly synchronize the compute stream before accessing the host memory.

Unlike the gather operation, where the output is always in device memory, host side synchronization is unnecessary.


linhu-nv (approved):
looks good to me.

PR #79, opened 2024-11-25
add devcontainers

jameslamb:
Follow-up to these PRs:

* https://github.com/rapidsai/devcontainers/pull/417
* https://github.com/rapidsai/cugraph-gnn/pull/68

Proposes adding devcontainers and a devcontainers CI job to the repo.

### Benefits of these changes

* faster and easier local development
* reduced risk of changes here breaking the RAPIDS unified devcontainers maintained in https://github.com/rapidsai/devcontainers

Similar to https://github.com/rapidsai/nx-cugraph/pull/25

### How I made these changes

Copied the `.devcontainer/` directory from https://github.com/rapidsai/cugraph, then just changed `cugraph` references to `cugraph-gnn`.

### How I tested this

Tested the `update-version.sh` changes like this:

```shell
./ci/release/update-version.sh '25.04.00'
git grep -E '25\.[0-9]+'
```


jameslamb:
Think this will need https://github.com/rapidsai/devcontainers/pull/423 before it can pass.

dependencies.yaml
```diff
@@ -262,6 +262,33 @@ dependencies:
         packages:
           - &cmake_ver cmake>=3.26.4,!=3.30.0
           - ninja
+    specific:
```
jameslamb, line 265: Copied this compiler block directly from `cugraph`: 

To try to fix this error in the conda devcontainer build:

```text
  CMake Error at /home/coder/.conda/envs/rapids/share/cmake-3.31/Modules/CMakeDetermineCXXCompiler.cmake:48 (message):
    Could not find compiler set in environment variable CXX:
  
    /usr/bin/g++.
  
  Call Stack (most recent call first):
    CMakeLists.txt:39 (project)
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12018692566/job/33548472621?pr=79))

Most other RAPIDS repos I checked have something like this, and I found offline conversations where it was suggested that the conda devcontainer builds expect the conda environment to provide compilers.

.devcontainer/cuda11.8-pip/devcontainer.json
```diff
@@ -0,0 +1,49 @@
+{
+  "build": {
+    "context": "${localWorkspaceFolder}/.devcontainer",
+    "dockerfile": "${localWorkspaceFolder}/.devcontainer/Dockerfile",
+    "args": {
+      "CUDA": "11.8",
+      "PYTHON_PACKAGE_MANAGER": "pip",
+      "BASE": "rapidsai/devcontainers:25.02-cpp-cuda11.8-ucx1.15.0-openmpi-ubuntu22.04"
```
bdice, line 8: This uses ucx 1.15 but the CUDA 12 container uses ucx 1.17. Is that right?

dependencies.yaml
bdice: Yes, this is a good fix.

bdice (approved):
One question about ucx versions. Otherwise approved.

.devcontainer/cuda11.8-pip/devcontainer.json
jameslamb: That's a great catch! I think it should be 1.17 here too.

I copied these directly from `cugraph`, looks like that was an oversight in https://github.com/rapidsai/cugraph/pull/4562#issuecomment-2258608174.

Let's try 1.17 here.

bdice: We need to fix this in cugraph and nx-cugraph. https://github.com/search?q=org%3Arapidsai%20ucx1.15&type=code

jameslamb:
wholegraph wheel builds are failing like this:

```text
  /__w/cugraph-gnn/cugraph-gnn/python/pylibwholegraph/build/cp311-cp311-linux_aarch64/_deps/raft-src/cpp/include/raft/core/detail/callback_sink.hpp:22:10: fatal error: spdlog/common.h: No such file or directory
     22 | #include <spdlog/common.h>
        |          ^~~~~~~~~~~~~~~~~
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12055077633/job/33614676583?pr=79#step:10:519))

I think that's a result of the ongoing logging changes, and should be fixed by https://github.com/rapidsai/raft/pull/2513.

.devcontainer/cuda11.8-pip/devcontainer.json
jameslamb: I just pushed a commit adding those changes (setting `RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true` in the pip devcontainers) here.

And manually cancelled CI because it'll fail until https://github.com/rapidsai/raft/pull/2513 is done.

PR #80, opened 2024-11-26
Require approval to run CI on draft PRs

bdice:
By default, CI runs on draft PRs. This leads to many CI runs that may be unnecessary.

With this PR's change to `.github/copy-pr-bot.yaml`, an `/ok to test` comment from a trusted user is required to trigger CI on draft PRs. Non-draft PRs will run CI by default, assuming that all commits are signed by trusted users. Otherwise an `/ok to test` is required (as before) -- see the `copy-pr-bot` docs at https://docs.gha-runners.nvidia.com/apps/copy-pr-bot/ for more information.

Part of https://github.com/rapidsai/build-planning/issues/123.

### DIFF
--- a/.github/copy-pr-bot.yaml
+++ b/.github/copy-pr-bot.yaml
@@ -2,3 +2,4 @@
 # https://docs.gha-runners.nvidia.com/apps/copy-pr-bot/
 
 enabled: true
+auto_sync_draft: false


jameslamb (approved):

bdice:
```
/pyenv/versions/3.10.15/lib/python3.10/site-packages/cudf/_lib/__init__.py:4: in <module>
    from . import (
parquet.pyx:1: in init cudf._lib.parquet
    ???
E   ValueError: pylibcudf.io.types.ColumnInMetadata size changed, may indicate binary incompatibility. Expected 112 from C header, got 48 from PyObject
```

These failures are unrelated. I suspect that two different builds of pylibcudf were used at different points in the build process. This should resolve itself, if I'm right. However, I don't think it is worth rerunning all CI for that to be tested, so I will request an admin-merge.

PR #82, opened 2024-12-02
[FEA] Support Heterogeneous Sampling in cuGraph-PyG

alexbarghi-nv:
Allows sampling of heterogeneous graphs.

Removes unbuffered sampling from the PyG examples and completely disables it in DGL.  A future PR will completely drop PyG support for unbuffered sampling, and a future `cugraph` PR will drop support for unbuffered sampling in the distributed sampler.

Merge after rapidsai/cugraph#4795

Closes rapidsai/cugraph#4402


python/cugraph-pyg/cugraph_pyg/tests/loader/test_neighbor_loader.py
```diff
@@ -194,3 +194,64 @@ def test_link_neighbor_loader_negative_sampling_uneven(batch_size):
...
+    feature_store = TensorDictFeatureStore()
+
+    graph_store[("paper", "cites", "paper"), "coo"] = [src, dst]
+    graph_store[("author", "writes", "paper"), "coo"] = [asrc, adst]
+
+    from cugraph_pyg.loader import NeighborLoader
+
+    loader = NeighborLoader(
+        (feature_store, graph_store),
+        num_neighbors=[0, 1, 0, 1],
```
tingyu66, line 248: Is the 0 fanouts here used to test edge cases?

tingyu66 (approved):

python/cugraph-pyg/cugraph_pyg/tests/loader/test_neighbor_loader.py
alexbarghi-nv: It's used to test whether we can properly exclude an edge type.

ci/test_notebooks.sh
```diff
@@ -11,7 +11,10 @@ rapids-logger "Generate notebook testing dependencies"
 rapids-dependency-file-generator \
   --output conda \
   --file-key test_notebooks \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee env.yaml
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
+  --prepend-channel "${CPP_CHANNEL}" \
+  --prepend-channel "${PYTHON_CHANNEL}" \
+| tee env.yaml
```
jameslamb, line 17: At this point in this script, `CPP_CHANNEL` and `PYTHON_CHANNEL` haven't yet been set. If you want the downloaded CI artifacts to be considered in the conda solve, you'll have to move this block from lower down up above this:

```shell
rapids-logger "Downloading artifacts from previous jobs"
CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
```

If you do that, then it'd also be good to remove the `rapids-mamba-retry install cugraph-dgl` in favor of that coming through this, so there will be a single call to create the environment.

Here's an example: https://github.com/rapidsai/ucx-py/pull/1101

dependencies.yaml
```diff
@@ -54,6 +54,8 @@ files:
     includes:
       - cuda_version
       - depends_on_pytorch
+      - depends_on_cudf
+      - depends_on_cugraph
```
jameslamb, line 58: Instead of this, I think here you want a `depends_on_cugraph_dgl`, so that a requirement like this gets added to the `env.yaml`:

```yaml
  - cugraph-dgl==25.2.*,>=0.0.0a0
```

Then `cudf` and `cugraph` will come through automatically as part of `cugraph-dgl`'s required dependencies.

https://github.com/rapidsai/cugraph-gnn/blob/87455cfedcc6721f24c783ba555af14a9a180624/conda/recipes/cugraph-dgl/meta.yaml#L27

That's a better pattern for CI, because it allows us to catch packaging problems of the form "`cugraph-dgl` depends on `cudf` but doesn't explicitly declare it" or something like that.

`cuspatial` uses this "consolidated solves" approach, you could follow that project's example:

https://github.com/rapidsai/cuspatial/blob/dbdc75ddea8422c18441a63e9f1fc42230db301c/dependencies.yaml#L43-L52

https://github.com/rapidsai/cuspatial/blob/dbdc75ddea8422c18441a63e9f1fc42230db301c/ci/test_notebooks.sh#L8-L19

jameslamb (commented):
Put up some suggestions on the notebook testing.

dependencies.yaml
```diff
@@ -54,6 +54,7 @@ files:
     includes:
       - cuda_version
       - depends_on_pytorch
+      - depends_on_cugraph_dgl
```
jameslamb, line 57: This won't work as-is because this project's `dependencies.yaml` doesn't yet have a `depends_on_cugraph_dgl`.

Add this:

```yaml
  depends_on_cugraph_dlg:
    common:
      - output_types: conda
        packages:
          - cugraph-dgl==25.2.*,>=0.0.0a0
```

Maybe here, after `depends_on_cugraph`:

https://github.com/alexbarghi-nv/cugraph-gnn/blob/9b19ee4b5407706dffc86ce971673133b33c63a4/dependencies.yaml#L543

It doesn't have to have as much stuff as `depends_on_cugraph` (for example), because we're only using it to reference conda packages.

Alternatively, you could avoid this `depends_on_cugraph_dgl` stuff (since this is the only reference) and instead add an item to the `test_notebook:` group (https://github.com/alexbarghi-nv/cugraph-gnn/blob/9b19ee4b5407706dffc86ce971673133b33c63a4/dependencies.yaml#L365C1-L372C16), so it'd look like this:

```yaml
  test_notebook:
    common:
      - output_types: [conda, requirements]
        packages:
          - ipython
          - nbconvert
          - notebook>=0.5.0
          - ogb
      - output_types: [conda]
        packages:
          - cugraph-dgl==25.2.*,>=0.0.0a0
```

alexbarghi-nv: Should be fixed now.  Thanks for the great explanation @jameslamb !

jameslamb: no prob, happy to help 😊

ci/test_notebooks.sh
```diff
@@ -1,17 +1,24 @@
...
+rapids-logger "Downloading artifacts from previous jobs"
+CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
+PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
+
 rapids-logger "Generate notebook testing dependencies"
 rapids-dependency-file-generator \
   --output conda \
   --file-key test_notebooks \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee env.yaml
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
```
jameslamb, line 18:
```suggestion
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
  --prepend-channel dglteam/label/th23_cu118 \
```

blegh sorry @alexbarghi-nv , I'd missed this when reviewing before... to get the `dgl` packages you want, I think you'll also need to add this channel to the `rapids-dependency-file-generator` call (as it was added in a `--channel` arg in the `rapids-mamba-retry install` below that we're removing.

Otherwise, I think this `dgl` requirement won't be met:

https://github.com/rapidsai/cugraph-gnn/blob/87455cfedcc6721f24c783ba555af14a9a180624/conda/recipes/cugraph-dgl/meta.yaml#L28

Because the only `dgl` that `mamba` will find is from conda-forge, which doesn't have `dgl>=2.4.0` or versions with stuff like `.th23`

image

https://anaconda.org/conda-forge/dgl/files

jameslamb (approved):
Seeing the notebook jobs pass! ([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12812235499/job/35724503000))

The only failing just is `cugraph-dgl` wheels testing, but that looks like a network error that'd be resolved by just re-running the job. 

> pip._vendor.urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='download.pytorch.org', port=443): Read timed out.

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12812235499/job/35724005309))

I just restarted it. This should be good to merge once that passes.

Thanks for getting this fixed so quickly @alexbarghi-nv !

jakirkham:
Thanks all! 🙏

PR #96, opened 2024-12-16
build libwholegraph docs in CI

### LINKED ISSUE #94: XML Docs for wholegraph needs to be created and uploaded in ci for use by cugraph-docs repository
This was done prior to migration into gnn

https://github.com/rapidsai/wholegraph/blob/branch-24.12/ci/build_docs.sh

jameslamb:
Fixes #94

Uploads API docs for `libwholegraph`, to be used by https://github.com/rapidsai/cugraph-docs/pull/46

Also removes `sphinx` dependencies... this repo only needs to produce Doxygen docs for `libwholegraph`, all the other Sphinx stuff will be done in https://github.com/rapidsai/cugraph-docs.


jameslamb:
Looks like this is working as expected:

> upload: ../../../tmp/tmp.M6YZWJmdr3/libwholegraph/xml_tar/xml.tar.gz to s3://rapids-downloads/ci/cugraph-gnn/pull-request/96/ef0cedc/docs/libwholegraph/xml_tar/xml.tar.gz

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12360403693/job/34495642435?pr=96#step:10:3299))

acostadon (approved):
Looks great

.github/workflows/build.yaml
```diff
@@ -43,6 +43,18 @@ jobs:
...
       sha: ${{ inputs.sha }}
+  docs-build:
+    needs: conda-cpp-build
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.02
+    with:
+      arch: "amd64"
+      branch: ${{ inputs.branch }}
+      build_type: ${{ inputs.build_type || 'branch' }}
+      container_image: "rapidsai/ci-conda:cuda12.5.1-ubuntu22.04-py3.11"
```
bdice, line 54: Can we use the `latest` tag? I think it's an oversight that we are not currently using `rapidsai/ci-conda:latest` for all our docs builds. We switched to a specific tag for a short time but now we should be able to use `latest` again. (Don't block on this if this PR is holding up other work.)

bdice (approved):
One suggestion to use `rapidsai/ci-conda:latest` -- but otherwise approving.

.github/workflows/build.yaml
jameslamb: `cugraph` actually is using `:latest` and I intentionally chose a specific tag here (though should have picked Python 3.12, that's an oversight).

I was thinking that a specific tag is helpful to be improve the probability that a branch is buildable for hotfixes even a few weeks after a release, if `ci-conda:latest` by that time points to a CUDA version or operating system or Python version that that release didn't support.

The tradeoff is that you get things like my mistake here, where the image being used isn't the latest one that the library could support.

Given that, do you still think `:latest`?

> *Don't block on this if this PR is holding up other work.*

Yeah we need this PR in to proceed with `cugraph-docs`, so I'll merge it whenever the next CI builds passes and put up a follow-up with whatever consensus we reach on the image tag.

bdice: It doesn’t really matter. If you want to pin it, that is fine, but we just need to be responsible for keeping those pins up to date. We also drop versions all the time (we may not need 12.5.1 in our test matrix at some future time) which can break this without warning. There is always a latest tag.

jameslamb: Yeah fair point. Changed my mind about waiting for a follow-up PR... jobs weren't actually that far along yet so I just pushed this change here: https://github.com/rapidsai/cugraph-gnn/pull/96/commits/5df836befc08bb75a522d2a51aec739a32d982a0

Issue #81, opened 2024-11-27
Remove dependency on libcugraphops / pylibcugraphops


jameslamb:
## Description

Starting in the 25.02 release, projects produced from this repo should not depend on any `cugraph-ops` packages.

## Benefits of this work

Faster CI.

Reduced effort for updates like adding new Python or CUDA versions.

## Acceptance Criteria

* `cugraph-pyg` and `cugraph-dgl` do not depend on any cugraph-ops packages

## Approach

N/A

## Notes

N/A

jameslamb:
closed by #99, thanks @tingyu66 !

alexbarghi-nv:
Closed by #99

PR #99, opened 2024-12-19
remove dependency on cugraph-ops

tingyu66:
Address #81


dependencies.yaml
```diff
@@ -431,18 +424,16 @@ dependencies:
...
         matrices:
-          - matrix: {cuda: "12.1"}
-            packages:
-              - pytorch-cuda=12.1
-          - matrix: {cuda: "12.4"}
+          - matrix: {cuda: "12.*"}
             packages:
-              - pytorch-cuda=12.4
-          - matrix: {cuda: "11.8"}
+            - pytorch-gpu>=2.3=*cuda120*
```
bdice, line 433: This is already using conda-forge, I think? `pytorch-gpu` is a `conda-forge` package, not a `pytorch` channel package. Also, the latest builds are built with CUDA 12.6. CUDA 12.0 is no longer used to build.

bdice: For compatibility reasons we may want to stick to older builds of pytorch-gpu (built with `cuda120`) for now. We will hopefully be able to relax this in the future.

tingyu66: Yes, this PR switches to `conda-forge::pytorch-gpu` since `pytorch` channel will discontinue.

> Also, the latest conda-forge builds are built with CUDA 12.6. CUDA 12.0 is no longer used to build.

> For compatibility reasons we may want to stick to older builds of pytorch-gpu (built with cuda120) for now. We will hopefully be able to relax this in the future.

Oh, I had not noticed that the most recent build (_306) is only against 12.6. I agree with keeping 12.0 for better backward compatibility. However, the CUDA 11 build seems missing. Do we have details on their build matrix?

bdice: CUDA 11 builds were dropped recently. You may need an older version for CUDA 11 compatibility. I also saw this while working on https://github.com/rapidsai/cudf/pull/17475. `mamba search -c conda-forge "pytorch=*=cuda118*"` indicates the latest version with CUDA 11 support is 2.5.1 build 303. The latest is 2.5.1 build 306.

bdice: For completeness, the latest CUDA 12.0 build was also 2.5.1 build 303.

tingyu66: Got it, thanks. It shouldn't be a dealbreaker unless another test component ends up requiring a newer version of torch on CUDA 11 down the line.

python/cugraph-pyg/cugraph_pyg/nn/conv/__init__.py
```diff
@@ -11,18 +11,31 @@
...
+try:
+    import pylibcugraphops
+
+    HAVE_CUGRAPH_OPS = True
+except ImportError:
+    pass
+except Exception as e:
+    warnings.warn(f"Unexpected error while importing pylibcugraphops: {e}")
+
+if HAVE_CUGRAPH_OPS:
```
jameslamb, line 26: Why is this PR putting cugraph-ops stuff behind a conditional import instead of completely removing it?

Or, said another way.... how are you expecting that someone would have `cugraph-pyg=25.2.*` and `pylibcugraphops=25.2.*` installed together? Is this being left here so it's still possible to build and install `pylibcugraphops` from source?

I'm hoping we can **completely stop publishing `pylibcugraphops` packages** in the 25.02 release.

jameslamb (commented):
Hey @tingyu66 , I'd like to help keep this moving forward. Left some questions for your consideration.

dependencies.yaml
```diff
@@ -18,7 +18,6 @@ files:
       - depends_on_cugraph
       - depends_on_cudf
       - depends_on_dask_cudf
-      - depends_on_pylibcugraphops
```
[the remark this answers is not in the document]
tingyu66: Yes

python/cugraph-pyg/cugraph_pyg/nn/conv/__init__.py
tingyu66: The conditional import is there because we planned to migrate cugraph-ops to a new location, and the new package will possibly retain the same name. We understand that rapids won't release cugraph-ops in 25.02. It's more of a placeholder for the migrated package.

dependencies.yaml
bdice: `pytorch-gpu` requires `__cuda`, and is not installable on systems without a CUDA driver. This makes it impossible to resolve the conda environment needed for devcontainers jobs in CI, which are CPU-only.

Note: many CUDA packages, including RAPIDS, are explicitly designed _not_ to have `__cuda` as a run requirement, because it makes it impossible to install on a CPU node before using that environment on another system with a GPU.

It looks like if we just use `pytorch` instead of `pytorch-gpu`, we still get GPU builds:

### CUDA 11:
```
CONDA_OVERRIDE_CUDA="11.8" conda create -n test --dry-run pytorch
```
shows
```
pytorch  2.5.1  cuda118_py313h40cdc2d_303  conda-forge
```

### CUDA 12:
```
CONDA_OVERRIDE_CUDA="12.5" conda create -n test --dry-run pytorch
```
shows
```
pytorch  2.5.1  cuda126_py313hae2543e_306  conda-forge
```

This should be sufficient. Let's try using just `pytorch` instead of `pytorch-gpu` with specific CUDA build selectors.

bdice: There are two benefits here, if my proposal above works.
1. devcontainers CI job would get CPU-only builds, which should still be fine for builds
2. We don't need to specify CUDA versions, so this dependency doesn't have to be "specific" to CUDA 11/12

jameslamb: I agree, let's try with `"pytorch"` instead of `"pytorch-gpu"`.

That opens up a risk that there may be situations where the solver chooses a CPU-only version because of some conflict, but hopefully `cugraph-pyg` can detect that with `torch.cuda.is_available()` or similar and raise an informative error saying something like "if using conda, try 'conda install cugraph-pyg pytorch-gpu'".

> *We don't need to specify CUDA versions, so this dependency doesn't have to be "specific" to CUDA 11/12*

I looked into this today... we shouldn't have needed to specify CUDA versions in build strings for `pytorch-gpu` anyway, as long as we're pinning the `cuda-version` package somewhere (for example, in the `run:` dependencies of `cugraph`).

Looks like `pytorch-gpu` is `==` pinned to a specific `pytorch`.

Screenshot 2025-01-07 at 2 00 40 PM

And the `pytorch` CUDA builds all have `run:` dependencies on `cuda-version`.

Screenshot 2025-01-07 at 2 02 22 PM

So here in `cugraph-pyg`, just having `cuda-version` as a `run:` dependency would be enough to ensure a compatible `pytorch-gpu` / `pytorch` is pulled.

bdice: @jameslamb These simplifications to drop build string info are only possible now with conda-forge, iirc. I believe more complexity was required when we used the pytorch channel, and we probably just carried that over when switching to conda-forge.

python/cugraph-pyg/cugraph_pyg/nn/conv/__init__.py
jameslamb: Oh interesting, did not know that!

Ok thank you.

dependencies.yaml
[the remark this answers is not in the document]
jameslamb: @tingyu66 could you please make this change (removing the dependencies on `pylibcugraphops` in conda recipes) in this PR?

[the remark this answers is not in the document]
tingyu66: @jameslamb Done, thank you!

bdice (commented):
The packaging changes look mostly right to me. There are a few areas that may have been missed:
- `ci/release/update-version.sh` has references to `libcugraphops` and `pylibcugraphops`
- `readme_pages/CONTRIBUTING.md` says "The cugraph-ops package is the expection being a closed-source package." (Also probably a typo, expection -> exception)
- `python/cugraph-pyg/pytest.ini` has references to `cugraph_ops`, and several tests are marked with `cugraph_ops`

I would like to see those addressed (or responded to here) before approving.

The cugraph-gnn library changes can probably go further than shown in this PR. I don't think it makes sense to keep so many references to a library that no longer exists, even if you expect a new package. Maybe it's better to drop all cugraph-ops references and then add those code paths back when the new home of that code exists.

jameslamb:
> *I don't think it makes sense to keep so many references to a library that no longer exists, even if you expect a new package. Maybe it's better to drop all cugraph-ops references and then add those code paths back when the new home of that code exists.*

I'm + to this. I think it'd be better to fully remove the effectively-unreachable code here and then re-introduce it when the successor to `cugraph-ops` is available.

But I'm also ok with deferring that to later if it means we could get this PR merged sooner, as that would unblock a lot of other work for removing `cugraph-ops` across RAPIDS.

tingyu66:
> The cugraph-gnn library changes can probably go further than shown in this PR. I don't think it makes sense to keep so many references to a library that no longer exists, even if you expect a new package. Maybe it's better to drop all cugraph-ops references and then add those code paths back when the new home of that code exists.

@jameslamb @bdice I agree with the suggestion to remove the unsupported code path from the codebase for now and reintroduce it later if needed. I will make the change today.

jameslamb:
I suspect that test failures like this:

```text
Traceback (most recent call last):
  File "/__w/cugraph-gnn/cugraph-gnn/python/cugraph-pyg/cugraph_pyg/examples/gcn_dist_sg.py", line 228, in <module>
    train(epoch)
  File "/__w/cugraph-gnn/cugraph-gnn/python/cugraph-pyg/cugraph_pyg/examples/gcn_dist_sg.py", line 49, in train
    for i, batch in enumerate(train_loader):
  File "/pyenv/versions/3.10.16/lib/python3.10/site-packages/cugraph_pyg/sampler/sampler.py", line 60, in __next__
    next_sample = next(self.__output_iter)
  File "/pyenv/versions/3.10.16/lib/python3.10/site-packages/cugraph_pyg/sampler/sampler.py", line 210, in __next__
    out = self._decode(self.__raw_sample_data, self.__index)
  File "/pyenv/versions/3.10.16/lib/python3.10/site-packages/cugraph_pyg/sampler/sampler.py", line 431, in _decode
    return self.__decode_csc(raw_sample_data, index)
  File "/pyenv/versions/3.10.16/lib/python3.10/site-packages/cugraph_pyg/sampler/sampler.py", line 247, in __decode_csc
    major_offsets_end_incl = raw_sample_data["label_hop_offsets"][
IndexError: index 388 is out of bounds for dimension 0 with size 387
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12759831095/job/35564523160?pr=99#step:9:2981))

are a result of https://github.com/rapidsai/cugraph/pull/4795, and that CI in this repo will be blocked until `cugraph-pyg` adapts.

I saw the same test failures in an unrelated wholegraph PR yesterday (#109, [build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/12753641679/job/35546681760?pr=109#step:8:9817)).

python/cugraph-dgl/cugraph_dgl/tests/test_utils.py
```diff
@@ -179,6 +180,7 @@ def test_get_source_destination_range():
     assert output_d == expected_output
 
 
+@pytest.mark.skip(reason="Skipping due to missing cugraph-ops backend.")
 def test__create_homogeneous_cugraph_dgl_nn_sparse_graph():
```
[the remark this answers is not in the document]
jameslamb: This file has dropped out of the diff completely in recent commits, now that the approach has changed to "just remove the unused code": https://github.com/rapidsai/cugraph-gnn/pull/99#issuecomment-2587655322

Resolving this thread.

python/cugraph-pyg/cugraph_pyg/nn/conv/__init__.py
jameslamb: This entire file is now just being deleted: https://github.com/rapidsai/cugraph-gnn/pull/99#issuecomment-2587655322

Resolving this thread.

jameslamb (commented):
These changes look good to me as-is.

However, it looks like this would still leave `pylibwholegraph` with a hard runtime dependency on `pylibcugraphops`.

I pulled this branch today and looked around like this:

```shell
git grep -E -i 'cugraph.*ops'
```

https://github.com/rapidsai/cugraph-gnn/blob/a9ab8b40476cfd5ac115fa5085e0a9f3bfa4eb72/python/pylibwholegraph/pylibwholegraph/torch/cugraphops/gat_conv.py#L18-L19

https://github.com/rapidsai/cugraph-gnn/blob/a9ab8b40476cfd5ac115fa5085e0a9f3bfa4eb72/python/pylibwholegraph/pylibwholegraph/torch/cugraphops/sage_conv.py#L19-L20

Can that be addressed in this PR? Without removing that, we can't stop building and shipping `pylibcugraphops` packages.

And there are other places (like `cugraph-pyg` tests decorated `@pytest.mark.cugraph_ops`), that @bdice mentioned here https://github.com/rapidsai/cugraph-gnn/pull/99#pullrequestreview-2538489955 ... can all those things also be removed?

tingyu66:
> These changes look good to me as-is.
> 
> However, it looks like this would still leave `pylibwholegraph` with a hard runtime dependency on `pylibcugraphops`.
> 
> I pulled this branch today and looked around like this:
> 
> ```shell
> git grep -E -i 'cugraph.*ops'
> ```
> 
> https://github.com/rapidsai/cugraph-gnn/blob/a9ab8b40476cfd5ac115fa5085e0a9f3bfa4eb72/python/pylibwholegraph/pylibwholegraph/torch/cugraphops/gat_conv.py#L18-L19
> 
> https://github.com/rapidsai/cugraph-gnn/blob/a9ab8b40476cfd5ac115fa5085e0a9f3bfa4eb72/python/pylibwholegraph/pylibwholegraph/torch/cugraphops/sage_conv.py#L19-L20
> 
> Can that be addressed in this PR? Without removing that, we can't stop building and shipping `pylibcugraphops` packages.
> 
> And there are other places (like `cugraph-pyg` tests decorated `@pytest.mark.cugraph_ops`), that @bdice mentioned here [#99 (review)](https://github.com/rapidsai/cugraph-gnn/pull/99#pullrequestreview-2538489955) ... can all those things also be removed?

I will remove the usage of the pytest decorator. I didn't realize that `pylibwholegraph.torch.gnn_model` also includes the cugraph-ops code path. I'll address it once I’m back at my keyboard today.

jameslamb:
Thank you! Sorry for not noting it earlier. I recommend you look at everything matched by this:

```shell
git grep -E -i 'cugraph.*ops'
```

And see if it can be removed.

Also update this to latest `branch-25.02`, to pull in Alex's testing changes from #82

bdice:
CI is currently passing. This PR is already pretty large. If this is in a relatively good state, albeit partially incomplete, perhaps we could merge this and continue the remaining work in a follow-up PR.

tingyu66:
> CI is currently passing. This PR is already pretty large. If this is in a relatively good state, albeit partially incomplete, perhaps we could merge this and continue the remaining work in a follow-up PR.

@bdice While we're still waiting for the codeowner review, I went ahead and squeezed in the fix. Thanks!

```
-> git grep -E -i 'cugraph.*ops'

cugraph-dgl/cugraph_dgl/dataloading/utils/sampling_helpers.py:    # and int64 respectively. Since pylibcugraphops binding code doesn't
cugraph-pyg/cugraph_pyg/tests/sampler/test_sampler_utils.py:@pytest.mark.cugraph_ops
cugraph-pyg/cugraph_pyg/tests/sampler/test_sampler_utils.py:@pytest.mark.cugraph_ops
cugraph-pyg/cugraph_pyg/tests/sampler/test_sampler_utils_mg.py:@pytest.mark.cugraph_ops
cugraph-pyg/cugraph_pyg/tests/sampler/test_sampler_utils_mg.py:@pytest.mark.cugraph_ops
cugraph-pyg/pytest.ini:          cugraph_ops: Tests requiring cugraph-ops
```
@jameslamb Above are the only cugraph-ops occurrences. The first one is a helpful note. Regarding the pytest markers, I've lost track of why they're marked that way in the first place, as they do not call cugraph-ops internally. @alexbarghi-nv, are those `uniform_neighbor_sample` calls in these tests specific to cugraph-ops use cases?

alexbarghi-nv:
Originally they used cugraph-ops at the C++ level but it's been taken out now.  We can safely remove these markers.

jameslamb (approved):
Thank you so much! I think all of my suggestions have been addressed and that we should merge this.

@alexbarghi-nv it looks like your approval would count for the other codeowners groups that need to approve here.

bdice (approved):
Build changes look good! Thanks for all this cleanup work!

alexbarghi-nv (approved):

PR #102, opened 2025-01-02
simplify wholegraph CMake, other small building and testing changes

jameslamb:
Proposes some miscellaneous packaging cleanup:

* sets minimum CMake version to 3.26.4 everywhere, to match the rest of RAPIDS
* removes commented-out CMake code
* removes unnecessary variables throughout CMake code
  - *including consolidating version references to use `RAPIDS_VERSION` from https://github.com/rapidsai/cugraph-gnn/blob/af22a1271251dc6b02d91cd593ac32b504356b8d/rapids_config.cmake#L20*
* updates some `pre-commit` hooks to their latest versions

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2023-2024, NVIDIA CORPORATION.
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
 
@@ -7,6 +7,7 @@ rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
 RAPIDS_VERSION="$(rapids-version)"
+export RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
 
 rapids-dependency-file-generator \
   --output conda \
--- a/cpp/Doxyfile
+++ b/cpp/Doxyfile
@@ -38,7 +38,7 @@ PROJECT_NAME           = "WholeGraph C API"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 25.02
+PROJECT_NUMBER         = $(RAPIDS_VERSION_MAJOR_MINOR)
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
 # for a project that appears at the top of each page and should give viewer a
--- a/python/pylibwholegraph/CMakeLists.txt
+++ b/python/pylibwholegraph/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2018-2024, NVIDIA CORPORATION.
+# Copyright (c) 2018-2025, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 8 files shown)


ci/build_docs.sh
```diff
@@ -1,12 +1,13 @@
...
-# Copyright (c) 2023-2024, NVIDIA CORPORATION.
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
 
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
 RAPIDS_VERSION="$(rapids-version)"
+export RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
```
jameslamb, line 10: For re-use in `cpp/Doxyfile`

cpp/Doxyfile
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           = "WholeGraph C API"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 25.02
+PROJECT_NUMBER         = $(RAPIDS_VERSION_MAJOR_MINOR)
```
jameslamb, line 41: This is what other RAPIDS projects are doing, e.g. `rmm`: https://github.com/rapidsai/rmm/blob/8275ba8dc94a7fc63d9513acd81e3a2cf30a85d1/doxygen/Doxyfile#L41

One less hard-coded version is nice for:

* smaller diffs on updates when development switches to new versions
* simpler `update-version.sh`
* reduced risk of inconsistencies between docs and code

linhu-nv (approved):
seems good to me. thx

python/pylibwholegraph/CMakeLists.txt
```diff
@@ -17,23 +17,17 @@
...
-FetchContent_MakeAvailable(rapids-cmake)
 include(rapids-cmake)
 include(rapids-cuda)
 include(rapids-cpm)
 
 rapids_cuda_init_architectures(PYLIBWHOLEGRAPH)
 
-project(PYLIBWHOLEGRAPH VERSION ${WHOLEGRAPH_VERSION} LANGUAGES CXX CUDA)
+project(
+  PYLIBWHOLEGRAPH
```
bdice, line 27: We're a little inconsistent about this across repos but it might be nice to use lowercase `pylibwholegraph` here. However, it's fine if that is out-of-scope for this PR (and/or not important enough to change at all).

e.g. cuDF has `project(CUDF...)` but also `project(pylibcudf...)`, so our casing is not consistent even within a repo.

bdice (approved):

python/pylibwholegraph/CMakeLists.txt
jameslamb: Thanks, yeah I'd noticed that. Let's treat it as out-of-scope for this PR. I'd like to merge this to simplify things ahead of setting up `libwholegraph` wheels.

The project name flows into a bunch of other variables that CMake generates (see https://cmake.org/cmake/help/latest/command/project.html#command:project) and I'm not sure if those are case-sensitive, so changing the case could require other changes to CMake code here (and maybe in other places that build wholegraph from source, like RAPIDS devcontainers). I'd rather not work through that right now.

bdice: Yes, all those downstream uses are case sensitive. At some point we may want to strive for consistency across RAPIDS but it’s low priority - lots of work, risk of breakage, and not a lot of value.

PR #103, opened 2025-01-06
remove flake8, clang tools from wholegraph CMake

jameslamb:
`wholgraph`'s CMake has some configuration to run `flake8`, `clang-tidy`, and `clang-foramt` via CMake.

This proposes removing it.

### Risks to doing this?

I don't think any.

`flake8` and `clang-format` configs are unnecessary, as those are already run via `pre-commit` here:

https://github.com/rapidsai/cugraph-gnn/blob/71675d868589ff9f904197f729985de1555cb914/.pre-commit-config.yaml#L22

https://github.com/rapidsai/cugraph-gnn/blob/71675d868589ff9f904197f729985de1555cb914/.pre-commit-config.yaml#L37

The `clang-tidy` support must not actually be used today... it refers to a script that doesn't exist in this repo:

https://github.com/rapidsai/cugraph-gnn/blob/71675d868589ff9f904197f729985de1555cb914/cpp/cmake/CodeChecker.cmake#L42-L43

This code had been in the `wholegraph` repo for a while... it was added in June 2023 (https://github.com/rapidsai/wholegraph/pull/24) and then never modified again.

### Benefits of doing this?

Similar to #102, I'm putting up PRs like this because I'm planning to attempt to add `libwholegraph` wheels, and want to simplify the `wholegraph` / `pylibwholegraph` CMake as much as possible before doing that, to reduce the implementation and reviewing effort.


python/pylibwholegraph/CMakeLists.txt
```diff
@@ -127,9 +127,6 @@ message(VERBOSE "PYLIBWHOLEGRAPH: Enable detection of conda environment for depe
 ##############################################################################
 # - Compiler options ---------------------------------------------------------
 
-# this is needed for clang-tidy runs
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
```
jameslamb, line 132: This is just a cleanup, not a behavior change... this option is already set to `ON` earlier in the same file.

https://github.com/rapidsai/cugraph-gnn/blob/71675d868589ff9f904197f729985de1555cb914/python/pylibwholegraph/CMakeLists.txt#L43

linhu-nv (approved):
thanks for cleaning this up. seems good to me :)

jameslamb:
Thanks @linhu-nv ! I might have one or two more coming like this, appreciate your help and patience 😊

PR #104, opened 2025-01-06
Heterogeneous Link Prediction Example for cuGraph-PyG

alexbarghi-nv:
Adds a heterogeneous link prediction example for cuGraph-PyG that uses the Taobao dataset.  Loosely based on the Taobao example from the PyG repository.

Adds ability to specify fanout as a dictionary to better align with PyG API.

Fixes a bug where the number of negative samples was calculated incorrectly, causing additional unwanted negative samples to be generated.

Updates the negative sampling call to match the new behavior added in rapidsai/cugraph#4885

Merge after rapidsai/cugraph#4898


python/cugraph-pyg/cugraph_pyg/data/graph_store.py
alexbarghi-nv, line 1: This is only a correction to the type hints.  Method behavior is unchanged.

alexbarghi-nv:
Confirmed this is working on 8xH100

tingyu66 (approved):

dependencies.yaml
```diff
@@ -428,6 +435,35 @@ dependencies:
               - *tensordict
           - {matrix: null, packages: [*pytorch_pip, *tensordict]}
 
+  # Will remove this after snap-stanford/ogb#497 is resolved.
+  depends_on_ogb:
+    common:
+      - output_types: [conda]
+        packages:
+          - ogb
```
KyleFromNVIDIA, line 443: Why is there now an additional dependency on `ogb`? The point of this is that we need to limit `pytorch` to `<2.6` due to https://github.com/snap-stanford/ogb/issues/497, is the below line not enough to do that?

alexbarghi-nv: You're right.  There's no reason to redeclare the dependency.  I'll remove it.

alexbarghi-nv: Done.  I also added an explanatory comment.

dependencies.yaml
```diff
@@ -428,6 +435,34 @@ dependencies:
...
           - {matrix: null, packages: [*pytorch_pip, *tensordict]}
 
+  # Will remove this after snap-stanford/ogb#497 is resolved.
+  # Temporarily sets the max pytorch version to 2.5 for compatibility
+  # with ogb.
+  depends_on_ogb:
+    common:
+      - output_types: [conda]
+        packages:
+          - pytorch>=2.3,<2.6a0
```
bdice, line 445: I'm a bit confused, do we also need an upper bound pinning for pip packages? It appears to be reusing `*pytorch_pip` below, which won't have the `<2.6a0` upper bound.

KyleFromNVIDIA (approved):

PR #105, opened 2025-01-07
Remove invalid conditional

vyasr:
The nightly tests have been failing despite all runs succeeding because the workflow's logic for filtering out notebook runs is invalid. Examples: https://github.com/rapidsai/cugraph-gnn/actions/runs/12649473784, https://github.com/rapidsai/cugraph-gnn/actions/runs/12609514484. Hopefully this change is sufficient to get the nightly suite passing.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,6 @@ jobs:
   conda-notebook-tests:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.02
-    if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_notebooks
     with:
       build_type: pull-request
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
This looks like a good fix, thanks.

Wild that this mistake somehow... put GitHub Actions into a state that caused it throw an unhandled exception???

> [test](https://github.com/rapidsai/cugraph-gnn/actions/runs/12649473784/workflow)
GitHub Actions has encountered an internal error when running your job.

build link: https://github.com/rapidsai/cugraph-gnn/actions/runs/12649473784

vyasr:
Yeah I wasn't sure what caused that at first. TBH I'm still not 100% sure it is, maybe we'll merge this fix but still see errors... but this is definitely not correct so hopefully this is all that was the problem.

PR #109, opened 2025-01-13
wholegraph: remove debugging details from CMake

jameslamb:
Proposes some simplifications for `wholegraph` CMake:

* removes code adding timing information via `RULE_LAUNCH_COMPILE` and `RULE_LAUNCH_LINK`
  - *these are internal to `ctest`, per https://cmake.org/cmake/help/latest/prop_dir/RULE_LAUNCH_LINK.html*
* removes `find_package(Python)` and related code in `pylibwholegraph`
  - *this is already handled by `rapids_cython_init()`: https://github.com/rapidsai/cugraph-gnn/blob/87455cfedcc6721f24c783ba555af14a9a180624/python/pylibwholegraph/CMakeLists.txt#L119-L120*

### Benefits of doing this?

Similar to #102 and #103, I'm putting up PRs like this because I'm planning to attempt to add libwholegraph wheels, and want to simplify the wholegraph / pylibwholegraph CMake as much as possible before doing that, to reduce the implementation and reviewing effort.


jameslamb:
I don't the CI failures here are related to this PR's changes. See https://github.com/rapidsai/cugraph-gnn/pull/99#issuecomment-2590307625

linhu-nv (approved):
seems good to me.

PR #110, opened 2025-01-14
WIP: [DO NOT MERGE] test libraft and libcugraph wheels

jameslamb:
**this should never be merged**

just opening it to test these:

* https://github.com/rapidsai/cugraph/pull/4804
* https://github.com/rapidsai/raft/pull/2531


jameslamb:
I don't think the CI failures here are related to this PR's changes. See https://github.com/rapidsai/cugraph-gnn/pull/99#issuecomment-2590307625

Seeing enough things succeed to make me think that the `cugraph` and `raft` changes won't break the package here.

bdice:
I bumped this to use published libraft wheels and updated the cugraph commit (may require a rerun once those artifacts are actually available, they are still building).

jameslamb:
I've updated the cugraph commit again, triggering a new run.

jameslamb:
This has served its testing purpose and can be closed now.

PR #119, opened 2025-01-30
Migrate to NVKS for amd64 CI runners

bdice:
This migrates amd64 CI jobs (PRs and nightlies) to use L4 GPUs from the NVKS cluster.

xref: https://github.com/rapidsai/build-infra/issues/184


jameslamb (commented):
`cugraph-pyg` and `cugraph-dgl` wheel tests both failed with network errors from `pip install`.

`conda-python-tests` have been running for almost 3.5 hours.

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/13077855044/job/36495325185?pr=119))

This might be a legitimate problem with NVKS. The `conda-python-tests` download some multi-GB datasets on every run... maybe that's prohibitively slow in the networking setup on NVKS? For comparison, those jobs took about 1-1.5 hours each to run successfully 3 days ago: https://github.com/rapidsai/cugraph-gnn/actions/runs/13020630286/job/36331117091

bdice:
It seems like the internet is a bit slower (the job has to complete or fail before I can get enough logs to see for sure). Currently it seems to be hung at either `Converting graphs into PyG objects...` or `Saving...`.

vyasr:
Didn't we do some work to avoid downloading in the cuml test suite? Could we port those lessons over to here?

alexbarghi-nv:
It looks like it's timing out downloading files.  I think we may need to drop the example tests entirely if this is an issue, unless we have the ogb datasets stored somewhere accessible to the runners.

jameslamb (approved):
approved!

PR #128, opened 2025-02-05
Fix of create_node_classification_datasets

linhu-nv:
quick fix to create_node_classification function: use data_and_label dict as the parameter instead of pickle_data_path


alexbarghi-nv (approved):
👍

PR #132, opened 2025-02-05
Update CuPy Version to 13.2

### LINKED ISSUE #22: GNN Packages: Update to cupy 13.2
https://github.com/rapidsai/cugraph/issues/4465 resolves the issue where we loaded the wrong NCCL version.  By setting the requirement for cupy 13.2, we can also upgrade the PyTorch version.

Requires this PR: https://github.com/rapidsai/cugraph/pull/4615

alexbarghi-nv:
CuPy 13.2 released a major fix that resolved an ongoing issue with NCCL (https://github.com/rapidsai/cugraph/issues/4465).  At this point, we are relying on other constraints to enforce the CuPy version, but to be safe, we should explicitly upgrade it.

Closes #22

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -14,7 +14,7 @@ dependencies:
 - cudatoolkit
 - cudf==25.4.*,>=0.0.0a0
 - cugraph==25.4.*,>=0.0.0a0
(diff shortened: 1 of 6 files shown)


bdice (approved):
Thanks! FYI we will bump cupy again for all of RAPIDS after the next cupy release to ensure Blackwell and CUDA 12.8 are supported.

PR #141, opened 2025-02-11
[FEA] New MovieLens Example, Add Timing to Taobao Example

alexbarghi-nv:
Adds a new example based on the MovieLens dataset in PyG.  Also updates the Taobao example with better timing and balances the edge_label/edge_label_index tensors in case the edgelist is distributed unevenly.

Corrects a small bug in the `LinkLoader` that prevented complete specification of heterogeneous graph input.


python/cugraph-pyg/cugraph_pyg/examples/taobao_mnmg.py
```diff
@@ -501,23 +506,38 @@ def create_loader(data_l):
...
     best_val_auc = 0
     for epoch in range(1, args.epochs + 1):
+        train_start = perf_counter()
         print("Train")
         loss = train(model, optimizer, train_loader)
+        train_end = perf_counter()
 
         if global_rank == 0:
             print("Val")
+        val_start = perf_counter()
```
tingyu66, line 528: I think it's safer to synchronize before perf_counter. `train()` should already sync implicitly due to the backward call, but I am not sure about `test()`.

tingyu66 (approved):
LGTM. Just left a nitpick.

PR #144, opened 2025-02-11
Create Conda CI test env in one step

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-planning/issues/22

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022-2024, NVIDIA CORPORATION.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
 
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -6,6 +6,12 @@ set -euo pipefail
 # Support invoking test_python.sh outside the script directory
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 
+if [[ "${RAPIDS_CUDA_VERSION%%.*}" == "11" ]]; then
+  DGL_CHANNEL="dglteam/label/th23_cu118"
+else
+  DGL_CHANNEL="dglteam/label/th23_cu121"
+fi
+
 . /opt/conda/etc/profile.d/conda.sh
 
 RAPIDS_VERSION="$(rapids-version)"
@@ -41,7 +47,7 @@ set +e
 # bulk sampler IO tests (hangs in CI)
 
 if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
-  rapids-logger "Generate Python testing dependencies"
+  rapids-logger "(cugraph-dgl) Generate Python testing dependencies"
   rapids-dependency-file-generator \
     --output conda \
     --file-key test_cugraph_dgl \
@@ -60,12 +66,6 @@ if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
   set +u
   conda activate test_cugraph_dgl
   set -u
-
-  if [[ "${RAPIDS_CUDA_VERSION%%.*}" == "11" ]]; then
-    DGL_CHANNEL="dglteam/label/th23_cu118"
-  else
-    DGL_CHANNEL="dglteam/label/th23_cu121"
-  fi
 
 
   rapids-print-env
@@ -90,7 +90,7 @@ else
 fi
 
 if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
-  rapids-logger "Generate Python testing dependencies"
+  rapids-logger "(cugraph-pyg) Generate Python testing dependencies"
   rapids-dependency-file-generator \
     --output conda \
     --file-key test_cugraph_pyg \
@@ -129,7 +129,7 @@ else
 fi
 
 if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
-  rapids-logger "Generate Python testing dependencies"
+  rapids-logger "(pylibwholegraph) Generate Python testing dependencies"
   rapids-dependency-file-generator \
     --output conda \
     --file-key test_pylibwholegraph \
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -14,15 +14,6 @@ rapids-logger "Downloading artifacts from previous jobs"
 CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
 PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
 
-rapids-logger "Generate Python testing dependencies"
-rapids-dependency-file-generator \
-  --output conda \
-  --file-key test_python \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
-  --prepend-channel "${CPP_CHANNEL}" \
-  --prepend-channel "${PYTHON_CHANNEL}" \
-| tee env.yaml
-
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -393,6 +433,7 @@ dependencies:
       - output_types: [conda, pyproject, requirements]
         packages:
           - pytest-forked
+      - output_types: [pyproject, requirements]
           - scipy
```
msarahan, line 437: YAML SNAFU here. I don't understand why you have added the output_types here. It can't split the packages list, though.

KyleFromNVIDIA: I meant to add `packages:`, which I have now done. I split it because the Conda environment doesn't include `scipy`.

KyleFromNVIDIA: I see that `scipy` was supposed to be in the other environments, so I've put it back.

ci/test_python.sh
bdice, line 1: Are ARM tests doing anything here? Maybe they should exit earlier if not.

There are a few comments of the form `# Reactivate the test environment back`. I’m not sure if we have a test environment to go back to, with the new single-solve environments.

KyleFromNVIDIA: We can do that in a follow-up PR. I'd like to limit the scope of this one to consolidating the environment creation.

ci/test_python.sh
```diff
@@ -50,6 +41,19 @@ set +e
...
+  rapids-logger "Generate Python testing dependencies"
+  rapids-dependency-file-generator \
+    --output conda \
+    --file-key test_cugraph_dgl \
+    --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
+    --prepend-channel "${CPP_CHANNEL}" \
+    --prepend-channel "${PYTHON_CHANNEL}" \
+    --prepend-channel pytorch \
+    --prepend-channel conda-forge \
+    --prepend-channel "${DGL_CHANNEL}" \
```
jameslamb, line 53: `DGL_CHANNEL` is computed further below, in that block beginning

```shell
 if [[ "${RAPIDS_CUDA_VERSION%%.*}" == "11" ]]; then
```

That needs to be moved up above this.

ci/test_python.sh
```diff
@@ -50,6 +41,19 @@ set +e
 # bulk sampler IO tests (hangs in CI)
 
 if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
+  rapids-logger "Generate Python testing dependencies"
```
jameslamb, line 44:
```suggestion
  rapids-logger "(cugraph_dgl) Generate Python testing dependencies"
```

I think we should add something to these messages to help differentiate them.

ci/test_python.sh
```diff
@@ -98,22 +90,23 @@ else
 fi
 
 if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
+  rapids-logger "Generate Python testing dependencies"
```
jameslamb, line 93:
```suggestion
  rapids-logger "(cugraph_pyg) Generate Python testing dependencies"
```

ci/test_python.sh
```diff
@@ -136,23 +129,23 @@ else
...
```
jameslamb, line 132:
```suggestion
  rapids-logger "(pylibwholegraph) Generate Python testing dependencies"
```

ci/test_python.sh
```diff
@@ -136,23 +129,23 @@ else
...
+  rapids-dependency-file-generator \
+    --output conda \
+    --file-key test_pylibwholegraph \
+    --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}"  \
+    --prepend-channel "${CPP_CHANNEL}" \
+    --prepend-channel "${PYTHON_CHANNEL}" \
+    --prepend-channel pytorch \
```
jameslamb, line 139: Not a blocking comment for this PR, but @tingyu66 @alexbarghi-nv are we planning to drop use of the `pytorch` channel in the 25.04 release?

Linking this related conversation: https://github.com/rapidsai/cugraph-gnn/pull/99#discussion_r1892492252

dependencies.yaml
```diff
@@ -658,3 +713,13 @@ dependencies:
...
+  depends_on_cugraph_pyg:
+    common:
+      - output_types: conda
+        packages:
+          - cugraph-pyg==25.4.*,>=0.0.0a0
+  depends_on_mkl:
+    common:
+      - output_types: conda
+        packages:
+          - mkl<2024.1.0
```
jameslamb, line 725:
```suggestion
  depends_on_mkl:
    common:
      - output_types: conda
        packages:
          - mkl<2024.1.0
```

Since this is only used in one place, can you please add it to the existing `test_python_pylibwholegraph:` group instead of as a new `depends_on:` entry?

Context: https://github.com/rapidsai/cudf/pull/18014#discussion_r1956411255

dependencies.yaml
```diff
@@ -68,6 +71,46 @@ files:
...
+      - depends_on_cugraph_pyg
+  test_pylibwholegraph:
+    output: none
+    includes:
+      - cuda_version
+      - depends_on_cugraph
+      - depends_on_cudf
+      - depends_on_dgl
+      - depends_on_pytorch
+      - depends_on_ogb
```
jameslamb, line 108: I don't think this is what you want, to replace `conda install ogb`.

`depends_on_ogb`... doesn't seem to contain `ogb`:

https://github.com/rapidsai/cugraph-gnn/blob/2f8afdea8fcb46eae7b1a1d1ef74a37df5cb831d/dependencies.yaml#L438-L464

That looks like a mistake, not sure where it happened. @tingyu66 could you please take a look? Maybe that was the result of a bad merge conflict resolution or something. `depends_on_ogb` seems to just contain `torch` (which we already have `depends_on_pytorch` for)

jameslamb (changes requested):
Left some suggestions. Leaving a blocking review, as a couple should definitely be resolved before merging this, and the question about `depends_on_ogb` needs a little discussion.

dependencies.yaml
jameslamb: Ok fair enough, we can leave it as you have it.

dependencies.yaml
alexbarghi-nv: This is a temporary constraint of PyTorch that only applies to test environments where ogb is being used.

alexbarghi-nv: Bradley and I discussed this here: https://github.com/rapidsai/cugraph-gnn/pull/104

alexbarghi-nv: And on Slack

bdice: A better name for this might be `ogb_pytorch_constraint`.

jameslamb: Got it, thanks! The comment directly about this, about this being there for the benefit of `ogb`, now makes sense to me.

It still would be better to either name it `ogb_pytorch_constraint`, I think, but that doesn't have to hold up this PR.

ci/test_python.sh
jameslamb: Thank you! I can get that started in a follow-up PR if you'd like.

dependencies.yaml
alexbarghi-nv: Ok, also devcontainer builds will fail until #148 gets merged.  I expect that to pass CI now that the cudf fix is merged.

jameslamb: Now that that's merged, I've merged in latest `branch-25.04` here to re-run CI.

jameslamb (approved):
Approving, my comments have been addressed.

I'll do the renaming suggested in https://github.com/rapidsai/cugraph-gnn/pull/144#discussion_r1956784434 + dropping the `pytorch` channel in a follow-up PR.

PR #149, opened 2025-02-25
add cugraph-notebook-codeowners to CODEOWNERS

AyodeAwe:
Makes all `.ipynb` files reviewable by `@cugraph-notebook-codeowners`


jameslamb (approved):
Looks ok to me, just note this team still needs to be manually added as a collaborator in the Settings.

image

AyodeAwe:
It has been added. See https://github.com/rapidsai/ops/issues/3636#issuecomment-2518149996

I think that error might be a bug from Github's end. I've seen it a few times now

PR #148, opened 2025-02-16
Update DGL Packages to PyTorch 2.4

alexbarghi-nv:
Update the DGL packages to those using PyTorch 2.4, since the DGL 2.4 packages for PyTorch 2.3 are no longer available.  Should unblock CI.

PR #150, opened 2025-02-25
Use conda-build instead of conda-mambabuild

bdice:
This changes from `conda mambabuild` to `conda build`. Conda now uses the mamba solver so no performance regressions are expected.

This is a temporary change as we plan to migrate to `rattler-build` in the near future. However, this is needed sooner to drop `boa` and unblock Python 3.13 migrations.

xref: https://github.com/rapidsai/build-planning/issues/149

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -17,7 +17,7 @@ rapids-logger "Begin cpp build"
 
 sccache --zero-stats
 
-RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry mambabuild conda/recipes/libwholegraph
+RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry build conda/recipes/libwholegraph
 
 sccache --show-adv-stats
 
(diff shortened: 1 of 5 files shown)


jameslamb:
Notebook tests failed like this:

```text
LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides dgl >=2.4.0.th24.cu* needed by cugraph-dgl-25.04.00a30-py312_30

Could not solve for environment specs
The following package could not be installed
└─ cugraph-dgl 25.4.*,>=0.0.0a0  is not installable because it requires
   └─ dgl >=2.4.0.th24.cu* , which does not exist (perhaps a missing channel).
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/13530151292/job/37811134682?pr=150#step:10:70))

Looks to me like nightly tests have been failing the same way, so not related to the changes in this PR: https://github.com/rapidsai/cugraph-gnn/actions/runs/13516476677/job/37766082750

I suspect that there really is a missing channel... like maybe #148 and #144 passed CI separately but one was merged without updating to latest `branch-25.04` containing the changes from the other?

jameslamb (approved):
Approving so this can be merged whenever the notebook tests are fixed.

jameslamb (approved):
Re-approving, the updated channels look like the right fix to me.

tingyu66 (approved):

jakirkham (approved):

PR #155, opened 2025-03-06
Update PyTorch to use CUDA 12.6

### LINKED ISSUE #154: Update CUDA 12 Version in CI
We are currently using CUDA 12.1, which is being phased out by PyTorch.  We need to migrate to a later CUDA 12 version in this release.

alexbarghi-nv:
Updates the PyTorch test dependency to install the version for CUDA 12.6.  This is the latest CUDA version supported by PyTorch.  CUDA 12.1 support is being dropped.

Closes #154


tingyu66 (commented):
While looking at the [wheel test log](https://github.com/rapidsai/cugraph-gnn/actions/runs/13707332839/job/38336367638?pr=155#step:10:38), it appears that the `extra-index-url` for PyTorch was never picked up during `pip install cugraph_pyg_cu12[test]`. As a result, CI used the default PyTorch v2.6 on pypi which was built with 12.4 CTK. This hasn't caused any failures yet since we only perform wheel tests on CUDA 12.

raydouglass (approved):

Issue #159, opened 2025-03-17
[IMP] Drop TensorDictFeatureStore from cuGraph-PyG


alexbarghi-nv:
The `tensordict` package is creating a lot of additional work to get CI running.  In addition, it is not memory-efficient and does not align with the future development of cuGraph-PyG.  Dropping `TensorDictFeatureStore` is the first step to removing all use of `tensordict` and migrating to a WholeGraph-based alternative.

Drop `TensorDictFeatureStore` in 25.08

PR #160, opened 2025-03-17
Deprecate TensorDictFeatureStore in cuGraph-PyG

alexbarghi-nv:
First step in handling #159 .  The unified API will store all data in WholeGraph, which has superior performance and lower memory usage.  `TensorDictFeatureStore` is no longer a useful alternative.

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/data/__init__.py
+++ b/python/cugraph-pyg/cugraph_pyg/data/__init__.py
@@ -18,9 +18,18 @@ from cugraph_pyg.data.dask_graph_store import (
 )
 from cugraph_pyg.data.graph_store import GraphStore
 from cugraph_pyg.data.feature_store import (
-    TensorDictFeatureStore,
+    TensorDictFeatureStore as DEPRECATED__TensorDictFeatureStore,
     WholeFeatureStore,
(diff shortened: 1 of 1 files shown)


python/cugraph-pyg/cugraph_pyg/data/__init__.py
```diff
@@ -18,11 +18,20 @@
...
 
 
+def TensorDictFeatureStore(*args, **kwargs):
+    warnings.warn(
+        "TensorDictFeatureStore is deprecated.  Consider changing your "
+        "workflow to launch using 'torchrun' and store data in "
+        "the faster and more memory-efficient WholeFeatureStore instead.",
+        FutureWarning,
+    )
+
```
tingyu66, line 33: Missing a return class here.

tingyu66 (changes requested):

python/cugraph-pyg/cugraph_pyg/data/__init__.py
alexbarghi-nv: sorry, let me fix that

tingyu66 (approved):
:+1:

PR #163, opened 2025-03-21
Branch 25.06 forward resolve

alexbarghi-nv:
Resolves conflict blocking forward merge.  Also removes 11.4 tests from the `conda-python-tests` matrix.


alexbarghi-nv:
@jameslamb why are we testing on CUDA 11.4 again?  It looks like the filter has stopped working.

jameslamb:
There's now a CUDA 11.4 job in the `conda-python-tests` matrix for all RAPIDS repos: https://github.com/rapidsai/shared-workflows/pull/306

The only filter I see on `branch-25.06` in this repo is the one that says "only run `amd64` jobs"

https://github.com/rapidsai/cugraph-gnn/blob/5cfb2e8f7c4772347ef07d542399119b3132a10d/.github/workflows/pr.yaml#L122-L129

You'll have to modify that filter if you want to always skip those jobs here.

alexbarghi-nv:
@jameslamb can I get a a re-review here?

tingyu66 (approved):

PR #170, opened 2025-04-04
[HOTFIX] Add deprecation warnings for DGL classes

alexbarghi-nv:
Hotfix to release 25.04 that adds deprecation warnings for the dask API in cuGraph-DGL, which is no longer maintainable and needs to be removed by release 25.08.  Also adds a warning on import informing users that the cuGraph-DGL is no longer under active development and that we recommend users migrate to cuGraph-PyG.

### DIFF
--- a/python/cugraph-dgl/cugraph_dgl/__init__.py
+++ b/python/cugraph-dgl/cugraph_dgl/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2024, NVIDIA CORPORATION.
+# Copyright (c) 2019-2025, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


python/cugraph-dgl/cugraph_dgl/__init__.py
```diff
@@ -25,3 +26,19 @@
...
+        "and will be removed in release 25.08.",
+        FutureWarning,
+    )
+    return DEPRECATED__CuGraphStorage(*args, **kwargs)
+
+
+warnings.warn(
+    "cuGraph-DGL is no longer"
+    "under active development.  We strongly recommend migrating to"
+    "cuGraph-PyG."
```
tingyu66, line 43: Maybe mention "removed by release 25.08" here.

tingyu66 (approved):

python/cugraph-dgl/cugraph_dgl/__init__.py
alexbarghi-nv: It's not being removed in 25.08, so that would be false.  It's probably being removed in 25.12.  But everything depending on the bulk sampler will be removed in 25.08 and is marked as such.

tingyu66: Got it. I misread the PR description.

KyleFromNVIDIA (approved):

PR #172, opened 2025-04-08
[IMP] Make WholeGraph a Hard Dependency of cuGraph-PyG, Remove cuDF Spilling from Examples

alexbarghi-nv:
Ahead of the unified API PR, make WholeGraph a hard dependency of cuGraph-PyG.  Also remove `cudf` spilling since we are no longer using `cudf`.  Spilling can be accomplished by using managed memory with `rmm` which is what the examples have always done.


tingyu66 (commented):
Should we also propagate the change into `meta.yaml` as well?

alexbarghi-nv:
> Should we also propagate the change into `meta.yaml` as well?

@tingyu66 should be fixed now.

python/cugraph-pyg/cugraph_pyg/examples/gcn_dist_mnmg.py
```diff
@@ -59,10 +59,6 @@ def init_pytorch_worker(global_rank, local_rank, world_size, cugraph_id):
 
     cupy.cuda.set_allocator(rmm_cupy_allocator)
 
-    from cugraph.testing.mg_utils import enable_spilling
-
-    enable_spilling()
-
```
tingyu66, line 65: Sorry, I missed that initially, but it appears we've enabled spilling in every example. Could we remove all instances of it? LGTM otherwise

alexbarghi-nv: Done

raydouglass (approved):

bdice (approved):

Issue #167, opened 2025-03-27
[BUG] conda environment doesn't solve on arm64


trxcllnt:
Attempting to solve the conda environment on arm64 machines fails with this error:
```
PackagesNotFoundError: The following packages are not available from current channels:
  - dgl
```

I don't see any aarch64 files in [`dglteam`](https://anaconda.org/dglteam/dgl/files) on anaconda.org, so I'm curious how the aarch64 conda packages are working? Should the `dependencies.yaml` pins be updated to match `dgl >=2.4.0.th24.cu*` used in the conda recipe?

Alternatively, we can restrict this dependency to x86_64:
```diff
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -479,13 +479,13 @@ dependencies:
     specific:
       - output_types: [conda]
         matrices:
-          - matrix: {cuda: "12.*"}
+          - matrix: {arch: x86_64, cuda: "12.*"}
             packages:
               - "dglteam/label/th24_cu124::dgl"
-          - matrix: {cuda: "11.*"}
+          - matrix: {arch: x86_64, cuda: "11.*"}
             packages:
               - "dglteam/label/th24_cu118::dgl"
-          - {matrix: null, packages: ["dglteam/label/th24_cu124::dgl"]}
+          - {matrix: null, packages: null}
 
   depends_on_pyg:
     common:
```

alexbarghi-nv:
Right now we can't really solve anything on `arm64` because there are no appropriate packages for PyTorch.

alexbarghi-nv:
If you need to run `cugraph-pyg` or `wholegraph` with `arm64` we recommend the NGC containers.

trxcllnt:
@alexbarghi-nv I don't necessarily need to run anything, I just need the environment to solve so I can compile the library on ARM. Right now I can create pip environments, but conda environments don't because this package isn't restricted to x86.

alexbarghi-nv:
Ok, we need to restrict it to x86 then

PR #175, opened 2025-04-09
restrict dgl dependency to x86

trxcllnt:
Fixes https://github.com/rapidsai/cugraph-gnn/issues/167


gforsyth (commented):
Changes look good, but CI needs the updated conda environment files to be checked in

bdice (approved):

alexbarghi-nv (approved):
👍

PR #181, opened 2025-04-16
Modify CMakeLists to install major version

alexbarghi-nv:
Updates CMakeLists in `pylibwholegraph` to install an inexact/major version of `libwholegraph` which is needed for the recent hotfix to build.


linhu-nv (approved):
looks good. Thanks

PR #182, opened 2025-04-16
feat: add libwholegraph wheel

gforsyth:
Contributes to https://github.com/rapidsai/build-planning/issues/33

This PR proposes packaging `libwholegraph` as a wheel, which is then re-used by `pylibwholegraph`, `cugraph-dgl`, and `cugraph-pyg`.

It should avoid recompiling RAFT and RMM, as those are now both handled by their respective wheels.

The `wholegraph` library is now compiled as a shared library.


python/libwholegraph/libwholegraph/load.py
```diff
@@ -0,0 +1,99 @@
...
+
+import ctypes
+import os
+
+# Loading with RTLD_LOCAL adds the library itself to the loader's
+# loaded library cache without loading any symbols into the global
+# namespace. This allows libraries that express a dependency on
+# this library to be loaded later and successfully satisfy this dependency
+# without polluting the global symbol table with symbols from
+# libcugraph that could conflict with symbols from other DSOs.
```
jameslamb, line 24:
```suggestion
# libwholegraph that could conflict with symbols from other DSOs.
```

python/libwholegraph/libwholegraph/load.py
```diff
@@ -0,0 +1,99 @@
...
+        # 'libwholegraph' has a runtime dependency on 'libraft' et al.. However,
+        # that dependency might be satisfied by a conda package
+        # (which does not have any Python modules), instead of a wheel.
+        #
+        # In that situation, assume that a shared object is in a place where
+        # the loader can find it.
+        pass
+
+    prefer_system_installation = (
+        os.getenv("RAPIDS_LIBCUGRAPH_PREFER_SYSTEM_LIBRARY", "false").lower() != "false"
```
jameslamb, line 65:
```suggestion
        os.getenv("RAPIDS_LIBWHOLEGRAPH_PREFER_SYSTEM_LIBRARY", "false").lower() != "false"
```

python/libwholegraph/pyproject.toml
```diff
@@ -0,0 +1,90 @@
...
+# find_package(libwholegraph) can find the files from this wheel
+libwholegraph = "libwholegraph"
+
+[tool.pydistcheck]
+select = [
+    "distro-too-large-compressed",
+]
+
+# detect when package size grows significantly
+max_allowed_size_compressed = '1.2G'
```
jameslamb, line 63:
```suggestion
max_allowed_size_compressed = '0.4G'
```

This size threshold should be reduced to more tightly match the current wheel sizes.

Looking through the `build-wheel-libwholegraph` logs ([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/14577429557/job/40886595329?pr=182)), I saw stuff like this for both CUDA 11 and CUDA 12 wheels:

```text
file size
  * compressed size: 0.3G
  * uncompressed size: 0.6G
  * compression space saving: 45.4%
```

python/pylibwholegraph/pyproject.toml
```diff
@@ -55,9 +55,13 @@ test = [
 [tool.rapids-build-backend]
 build-backend = "scikit_build_core.build"
 dependencies-file = "../../dependencies.yaml"
+matrix-entry = "cuda_suffixed=true"
 requires = [
     "cmake>=3.30.4",
     "cython>=3.0.0",
+    "libraft==25.6.*,>=0.0.0a0",
+    "librmm==25.6.*,>=0.0.0a0",
+    "libwholegraph==25.6.*,>=0.0.0a0",
```
jameslamb, line 64: `pylibwholegraph` wheels should pick up a **runtime** dependency on `libwholegraph` wheels too, because you expect to be able to `import libwholegraph` in `pylibwholegraph/__init__.py`.

python/libwholegraph/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+rapids_cuda_init_architectures(LIBWHOLEGRAPH)
+
+project(
+  LIBWHOLEGRAPH
+  VERSION "${RAPIDS_VERSION}"
+  LANGUAGES CXX CUDA
+)
+
+rapids_cpm_init()
```
jameslamb, line 30:
```suggestion
include(../../cmake/rapids_config.cmake)
include(rapids-cmake)
include(rapids-cuda)

rapids_cuda_init_architectures(LIBWHOLEGRAPH)

project(
  LIBWHOLEGRAPH
  VERSION "${RAPIDS_VERSION}"
  LANGUAGES CXX CUDA
)
```

I think the CPM stuff can be removed here. `rapids_cpm_init()` will be called here, in the project you've added via `add_subdirectory()`:

https://github.com/rapidsai/cugraph-gnn/blob/3d67322d679c8c08e7bd973279db378dff483058/cpp/CMakeLists.txt#L113

You'd only need to call `rapids_cpm_init()` up here in `libwholegraph/CMakeLists.txt` if you were invoking any CPM codepaths directly from here, which you aren't.

ci/build_wheel.sh
```diff
@@ -5,23 +5,23 @@ set -euo pipefail
...
-        --exclude libcuda.so.1
-        --exclude libnvidia-ml.so.1
+#
+EXCLUDE_ARGS=(
+    --exclude libcuda.so.1
+    --exclude libnvidia-ml.so.1
+    --exclude librapids_logger.so
+    )
+
+if [[ "${package_name}" == "pylibwholegraph" ]]; then
```
jameslamb, line 20:
```suggestion
if [[ "${package_name}" != "libwholegraph" ]]; then
```

I think it'd be a little safer to change this condition from "if is exactly `pylibwholegraph`" to "if is anything other than `libwholegraph`", so avoid accidentally vendoring `libwholegraph.so` in other wheels in this project, as the project evolves.

For example, did you double-checked that `libwholegraph.so` isn't now getting vendored by `auditwheel repair` in `cugraph-pyg` / `cugraph-dgl` builds, given the recently-merged changes in https://github.com/rapidsai/cugraph-gnn/pull/172 ?

ci/build_wheel_libwholegraph.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+package_dir="python/libwholegraph"
+
+export SKBUILD_CMAKE_ARGS="-DBUILD_SHARED_LIBS=OFF;-DCMAKE_MESSAGE_LOG_LEVEL=VERBOSE;-DCUDA_STATIC_RUNTIME=ON"
```
jameslamb, line 8: Why is `BUILD_SHARED_LIBS=OFF` here?

I'd expect it to be set to `ON`, because we're trying to produce a shared library here, not a static one (docs: https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html).

https://github.com/rapidsai/cugraph-gnn/blob/3d67322d679c8c08e7bd973279db378dff483058/cpp/CMakeLists.txt#L37

jameslamb (commented):
Hope you don't mind me crashing in, but I was interested when I saw @alexbarghi-nv post that this was up. Put up some comments for your consideration.

Leaving a "comment" review so my review doesn't block merging in case you get approvals between now and the next time I'm able to look at this.

ci/build_wheel.sh
gforsyth: It's not getting vendored in right now, but I don't know if I'm up-to-date against the default branch.  And this seems like a sensible precaution no matter what!

ci/build_wheel_libwholegraph.sh
gforsyth: I think just cargo-culted in from elsewhere, I can probably set this to `ON` and then remove the changes to the `cpp/CMakeLists.txt` file.  I'll try that locally and push up the results!

.github/workflows/build.yaml
```diff
@@ -114,7 +114,18 @@ jobs:
       date: ${{ inputs.date }}
       package-name: cugraph-pyg
       package-type: python
+  wheel-build-libwholegraph:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.06
+    with:
+      build_type: pull-request
+      script: ci/build_wheel_libwholegraph.sh
+      package-type: python
```
bdice, line 123: This is a "cpp" wheel. xref libcudf wheels here: https://github.com/rapidsai/cudf/blob/b259cd58a5b1086c72465f6e2aaafe7ce5973a7d/.github/workflows/build.yaml#L96
```suggestion
      package-type: cpp
```

.github/workflows/pr.yaml
```diff
@@ -136,10 +137,21 @@ jobs:
...
       run_script: "ci/build_docs.sh"
-  wheel-build-pylibwholegraph:
+  wheel-build-libwholegraph:
     needs: checks
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.06
+    with:
+      build_type: pull-request
+      script: ci/build_wheel_libwholegraph.sh
+      package-type: python
```
bdice, line 147:
```suggestion
      package-type: cpp
```

ci/build_wheel.sh
```diff
@@ -5,23 +5,23 @@ set -euo pipefail
...
-  python/pylibwholegraph)
-    EXCLUDE_ARGS=(
-        --exclude libcuda.so.1
-        --exclude libnvidia-ml.so.1
+#
+EXCLUDE_ARGS=(
+    --exclude libcuda.so.1
+    --exclude libnvidia-ml.so.1
+    --exclude librapids_logger.so
+    )
```
bdice, line 18:
```suggestion
)
```

python/pylibwholegraph/pyproject.toml
gforsyth: Done

ci/build_wheel_libwholegraph.sh
gforsyth: Ok, pushed up and I'll keep an eye on the logs

ci/build_wheel.sh
```diff
@@ -56,5 +56,9 @@ else
         -w "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}" \
         dist/*
 
-    RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 python "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"
+    if [[ ${package_name} == "libwholegraph" ]]; then
```
bdice, line 59: cuGraph (and maybe other libraries) solve this differently, with a third argument `package_type`. I think that's a bit cleaner. See snippets linked below.

https://github.com/rapidsai/cugraph/blob/60a6990d9e5087e4579d5b462a6a48f40fa3d391/ci/build_wheel.sh#L8

https://github.com/rapidsai/cugraph/blob/60a6990d9e5087e4579d5b462a6a48f40fa3d391/ci/build_wheel_libcugraph.sh#L31

ci/build_wheel_pylibwholegraph.sh
```diff
@@ -5,7 +5,22 @@ set -euo pipefail
...
+# Using env variable PIP_CONSTRAINT is necessary to ensure the constraints
+# are used when creating the isolated build environment.
+export PIP_CONSTRAINT="/tmp/constraints.txt"
+
+export SKBUILD_CMAKE_ARGS="-DBUILD_SHARED_LIBS=ON;-DCMAKE_MESSAGE_LOG_LEVEL=VERBOSE;-DCUDA_STATIC_RUNTIME=ON;-DWHOLEGRAPH_BUILD_WHEELS=ON"
 
 ./ci/build_wheel.sh pylibwholegraph ${package_dir}
 ./ci/validate_wheel.sh ${package_dir} "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"
+
+set +x
```
bdice, line 26: Do we need this?

python/libwholegraph/libwholegraph/load.py
```diff
@@ -0,0 +1,99 @@
...
+    return None
+
+
+def load_library():
+    """Dynamically load libwholegraph.so and its dependencies"""
+    try:
+        # these libraries must be loaded before libwholegraph because
+        # libwholegraph references their symbols
+        import libraft
+        import rapids_logger
```
bdice, line 51: Just to check -- is RMM used by wholegraph? I don't think it is but I could be misremembering.

ci/build_wheel.sh
gforsyth: thanks for the links, updated

ci/build_wheel_pylibwholegraph.sh
gforsyth: nope, leftover from debugging, removed.

python/libwholegraph/libwholegraph/load.py
gforsyth: it's not a runtime dependency

bdice:
The pip devcontainer build is looking for `libwholegraph-cu12`, but it should just look for the unsuffixed package iirc. I think you need some kind of `cuda_suffixed: "true"` thing like we do here: https://github.com/rapidsai/cudf/blob/bd588a8949de004685ffbf953d817c003006a3e6/dependencies.yaml#L1020-L1033

Also you might need to update the devcontainers like we did here for cuml: https://github.com/rapidsai/devcontainers/pull/442

jameslamb:
> I think it's failing because there are no libwholegraph wheels until this gets merged.

That's not it. `libwholegraph` wheels should never be *downloaded* in `devcontainers` builds here. There's code in the `devcontainers` setup to install dependencies, and it filters out of that anything that it knows it's building directly from source. Here's where that happens: https://github.com/rapidsai/devcontainers/blob/branch-25.06/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-dependencies.sh#L182-L191

@bdice 's advice in https://github.com/rapidsai/cugraph-gnn/pull/182#issuecomment-2819909071 is the fix... get a `libwholegraph` entry into the `rapids-build-utils` manifest in the `devcontainers` repo, and that should fix the build here.

gforsyth:
devcontainers are currently broken by `ucxx` -- should that block this from going in?

bdice:
> devcontainers are currently broken by `ucxx` -- should that block this from going in?

Replied here: https://github.com/rapidsai/devcontainers/pull/494#discussion_r2054483514

alexbarghi-nv (approved):
Approved; mostly deferring to the ops team on the finer details of the build process.

bdice:
All looks fine here with one required follow-up: we're shipping a ~350MB wheel but [the conda package is only 18MB](https://anaconda.org/rapidsai/libwholegraph/files). The difference appears to be in NCCL. We need to stop shipping NCCL in this wheel as @divyegala recently did with [RAFT](https://github.com/rapidsai/raft/pull/2629) and [cuVS](https://github.com/rapidsai/cuvs/pull/827).

@gforsyth Here's what I propose:
- Merge https://github.com/rapidsai/devcontainers/pull/494 now
- Rerun pip devcontainers and merge this PR immediately afterward
- Create follow-up for NCCL devendoring

bdice (approved):

gforsyth:
Yeah, I have a note down for the NCCL devendoring followup

PR #193, opened 2025-04-30
fix(libwholegraph): set build_type to "nightly" and also publish wheel

gforsyth:
Followup and fixes for #182

I forget to update "build-type" when copying over the new job from `py.yaml` to
`build.yaml` so nightly jobs were looking for an artifact in the nightly bucket
path, but the artifacts were uploaded to the pr path.

Also added a wheel publish step which I missed in the first PR.


bdice (approved):

alexbarghi-nv:
@bdice I think this will need an admin merge

PR #201, opened 2025-05-08
chore: skip DGL tests on Python 3.13

gforsyth:
`dgl` doesn't have Python 3.13 builds and it's also going to be dropped in 25.08, so we just aren't including it in the 3.13 migration.

Part of issue: https://github.com/rapidsai/build-planning/issues/120


alexbarghi-nv:
I modified `depenencies.yaml` to deal with build issues where it still attempts to install DGL on Python 3.13

.github/workflows/build.yaml
```diff
@@ -68,8 +68,9 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda-12.9.0
     with:
-      # This selects "ARCH=amd64 + the latest supported Python + CUDA".
-      matrix_filter: map(select(.ARCH == "amd64")) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
+      # This selects "ARCH=amd64 + Python 3.12 + CUDA".
```
bdice, line 71: This would drop Python 3.10 and 3.11, which I don’t think we want? Just say “not 3.13” here.

ci/test_python.sh
```diff
@@ -46,7 +46,7 @@ set +e
 # FIXME: TEMPORARILY disable MG PropertyGraph tests (experimental) tests and
 # bulk sampler IO tests (hangs in CI)
 
-if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
+if [[ "${RUNNER_ARCH}" != "ARM64" || "$RAPIDS_PY_VERSION" != "3.13" ]]; then
```
bdice, line 49:
```suggestion
if [[ "${RUNNER_ARCH}" != "ARM64" || "${RAPIDS_PY_VERSION}" != "3.13" ]]; then
```

.github/workflows/build.yaml
bdice: Same in pr.yaml.

gforsyth: It was already dropping Python 3.10 and 3.11 with the `max_by(.PY_VER)` -- I'm just telling it to use 3.12 instead of the new max of 3.13

alexbarghi-nv (approved):
👍

Thanks @gforsyth and @bdice for getting this done

.github/workflows/build.yaml
bdice: Ahhhh I missed that this is a pure wheel. (See `pure-wheel: true` below). That's fine, then.

ci/build_wheel_cugraph-dgl.sh
```diff
@@ -3,7 +3,11 @@
 
 set -euo pipefail
 
-package_dir="python/cugraph-dgl"
+if [ "$RAPIDS_PY_VERSION" != "3.13" ]; then
```
bdice, line 6: Shouldn't this be unreachable, because we skip the wheel build job for Python 3.13?

We also skip ARM but we don't have exceptions in the script for that. I would favor just leaving this script as-is and letting the filtering be done at the GHA workflow level.

The same applies to all bash scripts in this PR.

ci/test_python.sh
```diff
@@ -46,7 +46,7 @@ set +e
 # FIXME: TEMPORARILY disable MG PropertyGraph tests (experimental) tests and
 # bulk sampler IO tests (hangs in CI)
 
-if [[ "${RUNNER_ARCH}" != "ARM64" ]]; then
+if [[ "${RUNNER_ARCH}" != "ARM64" && "${RAPIDS_PY_VERSION}" != "3.13" ]]; then
```
bdice, line 49: Can we skip these with a matrix filter instead? I think currently these spin up a job (for ARM) that immediately ends. Let’s avoid that.

ci/run_cugraph_dgl_pytests.sh
bdice, line 1: No material changes. Please revert this file.

bdice (approved):
@gforsyth I deleted the merge comment so that I can pre-emptively approve this. Please look at my final round of comments and trigger a merge when you are ready.

ci/test_wheel_cugraph-dgl.sh
```diff
@@ -17,9 +17,9 @@ CUGRAPH_DGL_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_
 PKG_CUDA_VER="$(echo ${CUDA_VERSION} | cut -d '.' -f1,2 | tr -d '.')"
 PKG_CUDA_VER_MAJOR=${PKG_CUDA_VER:0:2}
 if [[ "${PKG_CUDA_VER_MAJOR}" == "12" ]]; then
-  PYTORCH_CUDA_VER="121"
+PYTORCH_CUDA_VER="121"
 else
-  PYTORCH_CUDA_VER=$PKG_CUDA_VER
+PYTORCH_CUDA_VER=$PKG_CUDA_VER
 fi
```
jakirkham, line 23: It looks like this dedent change snuck in

jakirkham (commented):
Thanks Gil for working on this! Also thanks Bradley and Alex for reviewing 🙏

Commented on one small change below that caught my eye

PR #219, opened 2025-05-29
refactor(rattler): remove cuda11 options and general cleanup

gforsyth:
xref rapidsai/build-planning#184

### DIFF
--- a/conda/recipes/cugraph-pyg/conda_build_config.yaml
+++ b/conda/recipes/cugraph-pyg/conda_build_config.yaml
@@ -1,16 +1,3 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
-
-c_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-
-cxx_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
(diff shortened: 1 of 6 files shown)


conda/recipes/cugraph-pyg/conda_build_config.yaml
```diff
@@ -1,16 +1,3 @@
...
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-
-cxx_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-
-cuda_compiler:
-  - nvcc
-
 cmake_version:
```
vyasr, line 1: Can we remove this file entirely? I don't think we need any of it.

conda/recipes/cugraph-pyg/recipe.yaml
vyasr, line 1: It looks like we can also drop some of the context variables around cuda from this recipe as well as the ignoring of the cuda-version run export. It looks like there is no cuda here at all. I think cugraph-pyg is pure Python, right? So we should also be able to remove all compiler references in this file, as well as Cython.

vyasr (changes requested):

conda/recipes/cugraph-pyg/recipe.yaml
gforsyth: Removed the compiler references and `cython`, and removed the `cuda` context variables.

bdice (approved):

PR #224, opened 2025-06-03
Remove CUDA 11 from dependencies.yaml

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-planning/issues/184


dependencies.yaml
```diff
@@ -736,12 +652,9 @@ dependencies:
...
         matrices:
           - matrix: {cuda: "12.*"}
-            packages:
+            packages: &cupy_packages_cu12
               - cupy-cuda12x>=13.2.0
-          - matrix: {cuda: "11.*"}
-            packages: &cupy_packages_cu11
-              - cupy-cuda11x>=13.2.0
-          - {matrix: null, packages: *cupy_packages_cu11}
+          - {matrix: null, packages: *cupy_packages_cu12}
```
jakirkham, line 657: Amm noticing the two matrices are somewhat redundant as observed in this thread: https://github.com/rapidsai/nx-cugraph/pull/151#discussion_r2125075348

Let's discuss in that thread. When it resolves, we can update both threads with the outcome

jakirkham (commented):
Thanks Kyle! 🙏

This looks reasonable

Left one comment. Though let's discuss in the linked thread

dependencies.yaml
KyleFromNVIDIA: I went ahead and collapsed this as per https://github.com/rapidsai/nx-cugraph/pull/151#discussion_r2126865536.

alexbarghi-nv (approved):
👍

PR #228, opened 2025-06-17
Use GCC 14 in conda builds.

vyasr:
## Description
conda-forge is migrating to gcc 14, so this PR is updating for alignment.

See https://github.com/rapidsai/build-planning/issues/188


alexbarghi-nv (approved):
👍

vyasr:
I'm temporarily pinning the CMake version due to what appears to be a bug in the include directory detection. I'll open a follow-up PR unpinning to debug.

alexbarghi-nv (approved):
👍

PR #230, opened 2025-06-18
Temporarily Disable Bitcoin Example Tests

alexbarghi-nv:
Temporarily disables the bitcoin example tests due to ongoing SSL issues causing the tests to fail in CI.  Will eventually reenable these tests with the data hosted somewhere else.


bdice (approved):
Maybe add a comment explaining why these are disabled.

PR #233, opened 2025-06-30
gcc14-compatibility

linhu-nv:
g++14 doesn't have the function "std::sample()" anymore, I fix compatibility issue of std::sample for g++14 and later, so that we can call this function no matter its g++ version


linhu-nv:
Add @alexbarghi-nv  for viz, and @vyasr, I rebased on your pr 228, so I think I have included your pr here.

alexbarghi-nv (approved):
👍

alexbarghi-nv:
Approved.  I'm surprised it doesn't support `std::sample` since it seems to still be part of the C++ standard library.

vyasr (changes requested):
Something seems wrong with this diagnosis. `std::sample` certainly still exists in with the gcc14 compiler, here's [a compiler explorer example](https://godbolt.org/z/fadqfTnMP). I can try to reproduce the problem.

vyasr:
I took a quick look at this. The issue isn't that the sort and sample algorithms are no longer supported, it's that the `<algorithm>` header was no longer being transitively included in the relevant libwholegraph files and that resulted in the compile failures. The fix in this PR worked because the `compat.hpp` header introduced also includes `<algorithm>`. That is why you had to include that header in `append_unique_tests.cu` even though you weren't using the `std_sample_impl` function it defines.

I've pushed suitable fixes to #228 now, so I'll close this PR. Sorry for taking up some of your time, I've been swamped with other things and was just hoping to spread the load a bit. Thanks for looking into it!

linhu-nv:
@vyasr  thanks for the deep dive and explain, I didn't get this very clear before I tried to fix it. My bad. Although I also think it's wierd that g++14 remove the support for std::sample. Now everything makes sense. Thanks!

PR #241, opened 2025-07-15
Migrate cugraph gnn packages to cugraph-pyg

nv-rliu:
Part of https://github.com/rapidsai/cugraph/issues/4407


rlratzel:
Thanks, @nv-rliu !  Is this intended to be merged now or are there more files to migrate?  Just wondering since I was thinking there might also be test coverage for these files and possible updates to CI scripts to include those tests.  If so, should this be in Draft?

nv-rliu:
I think it should be ready but deferring to @alexbarghi-nv in case I missed any files!

PR #263, opened 2025-07-28
cugraph_gnn RAPIDS dependency now use the new rapids-cmake branch info

robertmaynard:
This is required to handle when we switch RAPIDS branching strategy


linhu-nv (commented):
LGTM, thanks

PR #210, opened 2025-05-19
Remove cuGraph-DGL

alexbarghi-nv:
Removes cuGraph-DGL permanently and stops all CI and package publishing related to DGL support in this repository.

PR #264, opened 2025-07-29
remove cugraph-dgl CODEOWNERS rule, skip most CI on CODEOWNERS-only PRs

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/203

Follow-up to #210 

* removes a `CODEOWNERS` rule referencing `cugraph-dgl/`
* updates `changed-files` so that future PRs only modifying `CODEOWNERS` won't trigger a full CI run

## Notes for reviewers

Found these like this:

```shell
git grep -i dgl
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,7 +6,6 @@
 /readme_pages/     @rapidsai/cugraph-doc-codeowners
 
 # python code owners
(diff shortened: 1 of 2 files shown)


alexbarghi-nv (approved):
👍

PR #265, opened 2025-07-29
[FEA] Set random state using PyTorch generator

alexbarghi-nv:
Sets the random state using the PyTorch generator.

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/loader/utils.py
+++ b/python/cugraph-pyg/cugraph_pyg/loader/utils.py
@@ -1,0 +1,28 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+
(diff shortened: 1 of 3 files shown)


python/cugraph-pyg/cugraph_pyg/loader/utils.py
```diff
@@ -0,0 +1,28 @@
...
+def generate_seed():
+    world_size = torch.distributed.get_world_size()
+    rank = torch.distributed.get_rank()
+    if rank == 0:
+        seed = torch.randint(
+            0, 2**64 - world_size - 1, (1,), dtype=torch.uint64, device="cuda"
+        )
+    else:
+        seed = torch.tensor([0], dtype=torch.uint64, device="cuda")
+    torch.distributed.broadcast(seed, src=0)
```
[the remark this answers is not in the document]
alexbarghi-nv: It will just do nothing if there is only 1 worker.

python/cugraph-pyg/cugraph_pyg/loader/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+import torch
+
+
+def generate_seed():
+    world_size = torch.distributed.get_world_size()
+    rank = torch.distributed.get_rank()
+    if rank == 0:
+        seed = torch.randint(
+            0, 2**63 - world_size - 1, (1,), dtype=torch.int64, device="cuda"
```
tingyu66, line 22: intmax = 2**63 -1, rankmax = world_size -1, so IMO, high = intmax - world_size + 1 (1 due to it being exclusive in torch's api)

tingyu66 (approved):
:+1:

python/cugraph-pyg/cugraph_pyg/loader/utils.py
alexbarghi-nv: I agree, this should be `2**63 - 1 - world_size`.  I'll change it.

PR #271, opened 2025-08-06
Remove cugraph Python library as a dependency

### LINKED ISSUE #249: Update GNN Code to Use New Comms Functions
Update all code across the GNN packages to use the comms functions from `pylibcugraph` post-migration.

rlratzel:
Closes #249


alexbarghi-nv (approved):
Looks good, I think we can move this out of draft.

jameslamb (changes requested):
Came here looking for this because I was excited to see the note in @alexbarghi-nv's T5T saying the `cugraph` dependency had been removed from `cugraph-pyg`.

I can help you with a `packaging-codeowners` review and finishing this up.

Please also remove the dependency from the `cugraph-pyg` conda packages:

https://github.com/rapidsai/cugraph-gnn/blob/0f1f8384d22d2fdaf27c219e6ed908af4178aec9/conda/recipes/cugraph-pyg/recipe.yaml#L32

It also looks to me like there are more places where `depends_on_cugraph` could be removed in `dependencies.yaml`. If my comment about the use of `import_optional` and `MissingModule` in tests is correct, then it looks to me like `depends_on_cugraph` should just fully be removed from `dependencies.yaml`.

Otherwise, I see at least 2 specific cases that should be removed.

It looks to me like maybe the entire `test_python` group is unused and could be removed:

https://github.com/rapidsai/cugraph-gnn/blob/0f1f8384d22d2fdaf27c219e6ed908af4178aec9/dependencies.yaml#L58-L67

And I don't think `pylibwholegraph` needs `cugraph` for its tests:

https://github.com/rapidsai/cugraph-gnn/blob/0f1f8384d22d2fdaf27c219e6ed908af4178aec9/dependencies.yaml#L87

python/cugraph-pyg/cugraph_pyg/utils/imports.py
```diff
@@ -30,3 +30,67 @@ def package_available(requirement: str) -> bool:
             return False
 
     return True
+
+
+class MissingModule:
```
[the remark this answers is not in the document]
alexbarghi-nv: We're still keeping cuGraph as a test dependency, at least for now.  We need the datasets and some algorithms for a few tests.

alexbarghi-nv:
@jameslamb can you re-review?  I removed the unused dependencies as well.

python/cugraph-pyg/cugraph_pyg/utils/imports.py
```diff
@@ -30,3 +30,67 @@ def package_available(requirement: str) -> bool:
...
+    import the "optional" module 'mod' and return the module object or object.
+    If the import raises ModuleNotFoundError, returns an instance of
+    default_mod_class.
+
+    This method was written to support importing "optional" dependencies so
+    code can be written to run even if the dependency is not installed.
+
+    Example
+    -------
+    >> from cugraph.utils import import_optional
```
jameslamb, line 63:
```suggestion
    >> from cugraph_pyg.utils.imports import import_optional
```

These docstrings copied from `cugraph` should be updated as well.

python/cugraph-pyg/cugraph_pyg/utils/imports.py
```diff
@@ -30,3 +30,67 @@ def package_available(requirement: str) -> bool:
...
+    -------
+    >> class CuDFFallback:
+    ..   def __init__(self, mod_name):
+    ..     assert mod_name == "cudf"
+    ..     warnings.warn("cudf could not be imported, using pandas instead!")
+    ..   def __getattr__(self, attr):
+    ..     import pandas
+    ..     return getattr(pandas, attr)
+    ...
+    >> from cugraph.utils import import_optional
```
jameslamb, line 81:
```suggestion
    >> from from cugraph_pyg.utils.imports import import_optional
```

python/cugraph-pyg/cugraph_pyg/loader/link_neighbor_loader.py
```diff
@@ -130,25 +127,6 @@ def __init__(
...
-            See cugraph.gnn.DistSampleWriter.
-        batches_per_partition: int (optional, default=256)
-            The number of batches per partition if writing samples to
-            disk.  Manually tuning this parameter is not recommended
-            but reducing it may help conserve GPU memory.
-            See cugraph.gnn.DistSampleWriter.
-        format: str (optional, default='parquet')
-            If writing samples to disk, they will be written in this
-            file format.
-            See cugraph.gnn.DistSampleWriter.
```
jameslamb, line 151: If you do something like this on this branch:

```shell
git grep -E 'cugraph\.gnn'
```

You'll see more docs references like "See cugraph.gnn.DistSampleWriter.".

Should those be updated to reference something like `cugraph_pyg.sampler.distributed_sampler.DistributedNeighborSampler`?

alexbarghi-nv:
@jameslamb I think I may have removed some of the dependency lists that are actually being used.  The problem is it's not clear which are being used when I look at `dependencies.yaml`.

python/cugraph-pyg/pyproject.toml
```diff
@@ -44,8 +44,10 @@ Documentation = "https://docs.rapids.ai/api/cugraph/stable/"
 
 [project.optional-dependencies]
 test = [
+    "cugraph==25.10.*,>=0.0.0a0",
     "cuml==25.10.*,>=0.0.0a0",
     "ogb",
+    "pylibcugraph==25.10.*,>=0.0.0a0",
```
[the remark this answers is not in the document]
alexbarghi-nv: Yes, we should definitely fix that.

python/cugraph-pyg/cugraph_pyg/loader/link_neighbor_loader.py
alexbarghi-nv: Should be fixed now.

dependencies.yaml
```diff
@@ -48,153 +49,6 @@ files:
...
-      - depends_on_cudf
-      - depends_on_pytorch
-      - py_version
-      - test_python_common
-      - depends_on_mkl
-      - depends_on_pylibwholegraph
-      - test_python_pylibwholegraph
-  py_build_libwholegraph:
-    output: pyproject
-    pyproject_dir: python/libwholegraph
```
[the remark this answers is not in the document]
alexbarghi-nv: I think this is fixed.

dependencies.yaml
```diff
@@ -55,43 +56,6 @@ files:
...
-      - depends_on_pytorch
-      - depends_on_cuml
-      - depends_on_ogb
-      - depends_on_sentence_transformers
-      - py_version
-      - test_python_common
-      - depends_on_mkl
-      - depends_on_pylibwholegraph
-      - depends_on_cugraph_pyg
-  test_pylibwholegraph:
```
jameslamb, line 83: It looks like this group is used right here:

https://github.com/rapidsai/cugraph-gnn/blob/9ecbc668fa376ab7398e8eef9053aecbe510ac91/ci/test_python.sh#L74

Removing this is why the tests are failing like this:

```text
./ci/run_pylibwholegraph_pytests.sh: line 9: pytest: command not found
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/16944869038/job/48024049237?pr=271#step:10:1307))

I think that `test_pylibwholegraph` and `test_cugraph_pyg` should be restored (but `depends_on_cugraph` should be removed from `test_pylibwholegraph` because it looks like `cugraph` is not needed when testing `pylibwholegraph`).

```shell
git grep cugraph python/pylibwholegraph/
# (empty)
```

jameslamb (approved):
Looks good! Thanks for doing this, really nice simplification to the RAPIDS dependency graph.

I'll put up a follow-up PR with the changes mentioned in https://github.com/rapidsai/cugraph-gnn/pull/271/files#r2274062952

PR #276, opened 2025-08-13
Reduce seeds even more, enable use_on_oom for pools

alexbarghi-nv:
Further reduces `local_seeds_per_call` in CI and sets `use_on_oom=True` for all pools in attempt to resolve the ongoing OOM issues in CI.


alexbarghi-nv:
This still isn't working...
At this point, I think I am going to disable all example tests.

PR #277, opened 2025-08-14
WIP: cugraph-pyg: add missed runtime dependencies, remove runtime dependencies from 'test' extra

jameslamb:
Follow-up from #271 

Explicitly declares runtime dependencies on libraries that `cugraph-pyg` unconditionally imports.

`cudf`:

https://github.com/rapidsai/cugraph-gnn/blob/b10f279cb855b2cfc52239d2d7973c01c92c2e5f/python/cugraph-pyg/cugraph_pyg/sampler/io/reader.py#L19

`cupy`:

https://github.com/rapidsai/cugraph-gnn/blob/b10f279cb855b2cfc52239d2d7973c01c92c2e5f/python/cugraph-pyg/cugraph_pyg/data/graph_store.py#L17

`packaging`:

https://github.com/rapidsai/cugraph-gnn/blob/b10f279cb855b2cfc52239d2d7973c01c92c2e5f/python/cugraph-pyg/cugraph_pyg/utils/imports.py#L14

Removes `pylibcugraph`, `pylibwholegraph`, and `torch-geometric` from `[test]` extra (these are already in `[project].dependencies`).

Fixes homepage URL in `cugraph-pyg` project metadata.

### What should we do with `torch`?

There are a mix of conditional and unconditional imports of `torch` in `cugraph-pyg`:

https://github.com/rapidsai/cugraph-gnn/blob/b10f279cb855b2cfc52239d2d7973c01c92c2e5f/python/cugraph-pyg/cugraph_pyg/tensor/dist_tensor.py#L25

https://github.com/rapidsai/cugraph-gnn/blob/b10f279cb855b2cfc52239d2d7973c01c92c2e5f/python/cugraph-pyg/cugraph_pyg/loader/utils.py#L14

Unconditional imports really should mean declaring a hard runtime dependency, but I vaguely recall that in the past, we intentionally did not declare a `torch` dependency in `cugraph-pyg` wheels. Am I right about that? @alexbarghi-nv can you remind me why we don't declare that dependency? If I'm right, should those unconditional imports be made conditional with `import_optional()`?

### Benefits of this

Reduces the risk of import issues when users install `cugraph-pyg` in environments that don't happen to have these dependencies from other things (similar to https://github.com/rapidsai/cuml/pull/6420).

Improves the chance of catching packaging issues in CI.


alexbarghi-nv:
It looks like a lot of mistakes made it through.  Thanks for catching these.

`cudf` should not be a runtime dependency.  The code you mentioned that uses it was mistakenly copied from cuGraph and should be deleted.  I'll do that in another PR.

`torch` should not be listed as a runtime dependency, only as a test dependency.  This is so containers can be built without PyTorch.  Every use of it should be with `import_optional`.  I'll also fix that in another PR.

alexbarghi-nv:
@jameslamb https://github.com/rapidsai/cugraph-gnn/pull/278

alexbarghi-nv:
@jameslamb see #276

python/cugraph-pyg/pyproject.toml
```diff
@@ -31,30 +31,29 @@ classifiers = [
...
     "pandas",
     "pylibcugraph==25.10.*,>=0.0.0a0",
     "pylibwholegraph==25.10.*,>=0.0.0a0",
     "torch-geometric>=2.5,<2.7",
 ] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
 
 [project.urls]
-Homepage = "https://github.com/rapidsai/cugraph"
+Homepage = "https://github.com/rapidsai/cugraph-gnn"
 Documentation = "https://docs.rapids.ai/api/cugraph/stable/"
```
bdice, line 45: Just a side note, we should probably update this. The title tag on this page right now is "`* NOTICE *`", for the DGL deprecation. That is currently very loud, both visually and semantically.

bdice (approved):

PR #279, opened 2025-08-14
[CI] Disable Example Tests

alexbarghi-nv:
The cuGraph-PyG examples cannot be reliably tested in CI due to memory constraints.  We will need to come up with a new strategy for testing/verifying these examples going forward.


jameslamb (approved):
Thanks for this

PR #280, opened 2025-08-21
feat: adding docs for examples

paulhendricks:
Adds documentation for running examples.

Currently references 25.10 container that does not yet exist. Can merge when container is available and can continue to add to documentation until then.


alexbarghi-nv (changes requested):
Looks good, just drop the PyG container reference

PR #303, opened 2025-09-10
[BUG] Mask out unwanted vertices during negative sampling

### LINKED ISSUE #304: [BUG] Triplet Heterogeneous Negative Sampling is Probably Broken
Because of the strict batching requirements of cuGraph-PyG, we are concatenating the destination vertices twice during triplet negative sampling.  I think this will break the expectation that `edge_label_index` can be broken up into src/dst batches and cause de-offsetting of the edge label index to fail.

I am not sure how to fix this at the moment.

alexbarghi-nv:
Masks out unwanted vertices during heterogeneous negative sampling, which was previously not being done.  This caused de-offsetting to fail and produce negative values for `edge_label_index`, which exposed the bug.  Anything sampled with these negatives edge would have been invalid anyways even if de-offsetting returned positive values.

Also fixes a bug affecting triplet sampling by concatenating from a random subset of src instead of dst.

Closes #304 
Partially resolves nvbug#5502562

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/sampler/sampler.py
+++ b/python/cugraph-pyg/cugraph_pyg/sampler/sampler.py
@@ -827,7 +827,7 @@ class BaseSampler:
             if neg_sampling.is_binary():
                 src, _ = neg_cat(src.cuda(), src_neg, self.__batch_size)
             else:
-                # triplet, cat rand subset of src to src so length is the
+                # triplet, cat random subset of src to src so length is the
                 # same; will result in the same set of unique vertices
                 scu = src.cuda()
                 per = torch.randint(


alexbarghi-nv:
Related: https://github.com/rapidsai/cugraph-gnn/issues/304

python/cugraph-pyg/cugraph_pyg/sampler/sampler_utils.py
```diff
@@ -91,22 +92,65 @@ def neg_sample(
...
+                torch.arange(num_src_nodes, dtype=torch.int64, device="cuda")
+                + graph_store._vertex_offsets[input_type[0]]
+            )
+
+        src_weight = torch.concat(
+            [src_weight, torch.zeros(num_dst_nodes, dtype=torch.int64, device="cuda")]
+        )
+        dst_weight = torch.concat(
+            [torch.zeros(num_src_nodes, dtype=torch.int64, device="cuda"), dst_weight]
+        )
```
paulhendricks, line 142: Looking here: https://github.com/rapidsai/cugraph/blob/branch-25.10/python/pylibcugraph/pylibcugraph/negative_sampling.pyx#L100

Should this be:

```python
        src_weight = torch.concat(
            [src_weight, torch.zeros(num_dst_nodes, dtype=torch.float32, device="cuda")]
        )
        dst_weight = torch.concat(
            [torch.zeros(num_src_nodes, dtype=torch.float32, device="cuda"), dst_weight]
        )
```

alexbarghi-nv: Actually they can be float or double.  I'll make sure they are aligned properly

tingyu66 (approved):

Issue #305, opened 2025-09-10
[BUG] `len(edges)` in `distributed_sampler.py` is always 2


paulhendricks:
Core code:

```python
    def sample_from_edges(
        self,
        edges: TensorType,
        *,
        batch_size: int = 16,
        random_state: int = 62,
        assume_equal_input_size: bool = False,
        input_id: Optional[TensorType] = None,
        input_label: Optional[TensorType] = None,
        metadata: Optional[Dict[str, Union[str, Tuple[str, str, str]]]] = None,
    ) -> Iterator[Tuple[Dict[str, "torch.Tensor"], int, int]]:
        """
        Performs sampling starting from seed edges.

        Parameters
        ----------
        edges: TensorType
            2 x (# edges) tensor of edges to sample from.
            Standard src/dst format.  This will be converted
            to a list of seed nodes.
        batch_size: int
            The size of each batch.

...

        batches_per_call = self._local_seeds_per_call // batch_size
        actual_seed_edges_per_call = batches_per_call * batch_size

        if input_id is None:
            input_id = torch.arange(len(edges), dtype=torch.int64, device="cpu")

        local_num_batches = int(ceil(num_seed_edges / batch_size))

```

```python
import torch

N = 5
edges = torch.randint(0, 100, (2, N))  # shape [2, N]
print("edges shape:", edges.shape)

print("len(edges):", len(edges))           # number of rows (2)
print("edges.shape[-1]:", edges.shape[-1]) # number of columns (N)
```

Suggested fix:

```python
        if input_id is None:
            input_id = torch.arange(edges.shape[-1], dtype=torch.int64, device="cpu")
```

alexbarghi-nv:
Fixed by #316

PR #306, opened 2025-09-10
fix: fixes len edges to use columns instead of always being 2

paulhendricks:
Relates to: https://github.com/rapidsai/cugraph-gnn/issues/305


alexbarghi-nv:
@paulhendricks can you please give me permission to push to your branch by allowing maintainers to push to PR branches in the PR config?

PR #328, opened 2025-10-07
[BUG] Use memory pool in movielens example

alexbarghi-nv:
The movielens example currently uses the `rmm_torch_allocator` by changing the default PyTorch allocator, which doesn't work in a multi-GPU environment.  This PR fixes the example by using a `torch.cuda.MemPool` instead, which is what the other examples now do.


tingyu66 (approved):
👍

PR #343, opened 2025-11-06
Remove inheritance from deprecated unary_function

bdice:
`std::unary_function` was deprecated in C++11 and removed in C++17. This removes the inheritance from that base class. It is not necessary for proper functionality.

This fixes a large number of warnings like these:
```
/home/coder/cugraph-gnn/cpp/src/wholememory_ops/register.hpp:18:35: warning: 'template<class _Arg, class _Result> struct std::unary_function' is deprecated [-Wdeprecated-declarations]
```


linhu-nv:
One quick question, should we also remove inheritance for `three_wmt_hash`? thanks @bdice

bdice:
@linhu-nv Can you approve based on the conversation above?

linhu-nv (approved):
Approved, thanks

PR #346, opened 2025-11-13
[IMP] Support more dtypes in the cuGraph-PyG FeatureStore

alexbarghi-nv:
Adds support for several more dtypes, including `float16` and `int16` which are supported by WholeGraph but were not exposed through the cuGraph-PyG API.

Despite it being listed as a dtype in WholeGraph, `bfloat16` does not seem to work properly, so it is not included in this PR.

`torch.bool` is removed since it was never supported by WholeGraph, and its inclusion was a mistake.  Attempting to use `torch.bool` would have resulted in an exception anyways, so I still consider its removal to be non-breaking.


alexbarghi-nv:
Rerunning due to download issue

PR #369, opened 2025-12-09
Use static linkage for CUDA runtime

bdice:
## Summary
- Set `CUDA_STATIC_RUNTIME=ON` by default in `cpp/CMakeLists.txt`
- Remove `cuda-cudart` from run requirements in conda recipes

With static linking of the CUDA runtime, the runtime is embedded in the binaries and the `cuda-cudart` package is not needed at runtime.

Part of https://github.com/rapidsai/build-planning/issues/235


linhu-nv (approved):
@alexbarghi-nv  sorry I missed this. It looks good to me. thanks

PR #372, opened 2025-12-12
Fix compiler warnings in scatter_op_impl_mapped.cu

bdice:
This PR fixes some warnings I encountered when compiling wholegraph.

- `#128-D: loop is not reachable` (line 34)
- `#940-D: missing return statement at end of non-void function` (line 35)

This aligns with the pattern used in `gather_op_impl_mapped.cu`.


linhu-nv (approved):
LGTM. Thanks for the fix!

PR #379, opened 2026-01-02
[FEA] Support Temporal Negative Sampling in cuGraph-PyG

alexbarghi-nv:
Supports temporal negative sampling in cuGraph-PyG.  Currently only works for edge-based temporal sampling, which is actually _not_ currently supported in native PyG.  Does not yet support node-based temporal negative sampling, because we do not support node-based temporal sampling in general (this would require changes to `libcugraph`).


alexbarghi-nv:
Replaced by #382

PR #381, opened 2026-01-05
build and test against CUDA 13.1.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/236

Tests that CI here will work with the changes from https://github.com/rapidsai/shared-workflows/pull/483,
switches CUDA 13 builds to CUDA 13.1.0 and adds some CUDA 13.1.0 test jobs.


jameslamb:
Thanks for starting CI on a couple of these @jakirkham , but anything requiring NCCL (like RAFT, cuML, and cuGraph) is going to be stuck until we resolve this:

* https://github.com/rapidsai/raft/pull/2912#discussion_r2665715971
* https://github.com/rapidsai/devcontainers/issues/644

PR #382, opened 2026-01-08
[FEA] Support Temporal Negative Sampling, Add Temporal Features to MovieLens Example

alexbarghi-nv:
Adds support for temporal negative sampling in cuGraph-PyG. Adds temporal features and temporal sampling to the movielens example.


python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -70,7 +71,7 @@ def __clear_graph(self):
         self.__graph = None
         self.__vertex_offsets = None
         self.__weight_attr = None
-        self.__etime_attr = None
+        self.__time_attr = None
         self.__numeric_edge_types = None
```
tingyu66: I just found it a bit unconventional that these members are not defined in `__init__`. (I understand __clear_graph as a pure reset function)

tingyu66 (approved):

PR #407, opened 2026-02-10
refactor: build wheels and conda packages using Python limited API

gforsyth:
Moving `pylibwholegraph` over to using the limited API as part of https://github.com/rapidsai/build-planning/issues/42

Ops-Bot-Merge-Barrier: true

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -34,7 +34,6 @@ rapids-logger "Building pylibwholegraph"
 # more info is available at
 # https://rattler.build/latest/tips_and_tricks/#using-sccache-or-ccache-with-rattler-build
 rattler-build build --recipe conda/recipes/pylibwholegraph \
-                    --test skip \
                     "${RATTLER_ARGS[@]}" \
                     "${RATTLER_CHANNELS[@]}"
 


conda/recipes/pylibwholegraph/recipe.yaml
```diff
@@ -81,6 +86,12 @@ requirements:
     by_name:
       - cuda-version
 
+tests:
+  - python:
+      imports: pylibwholegraph
+      python_version: ${{ py_runtime_latest }}.*
+      pip_check: false
+
```
jameslamb: I wouldn't have expected this to work on a CPU-only runner, because I thought `pylibwholegraph` would need the GPU at runtime.

But I guess it does!

```console
$ docker run --rm -it rapidsai/ci-conda:26.04-latest bash
$ conda install --yes -c rapidsai-nightly pylibwholegraph
$ python -c "import pylibwholegraph; print(pylibwholegraph.__version__)"
26.04.00a18
```

That's great, import tests are helpful for catching issues! But... this isn't actually running in CI. You'll need to remove this:

https://github.com/rapidsai/cugraph-gnn/blob/2c04b3769ffec061d22861c039dfa8fa0ed206f2/ci/build_python.sh#L37

alexbarghi-nv (approved):
👍

PR #453, opened 2026-04-30
Use `token.rapids.nvidia.com` when issuing S3 bucket creds in devcontainers

trxcllnt:
Set AWS_IDP_URL and update AWS_ROLE_ARN to use `token.rapids.nvidia.com`


alexbarghi-nv (approved):
Looks good, thanks

PR #461, opened 2026-05-14
Update to 26.08

jolorunyomi:
This PR updates the repository to version 26.08.

This is part of the 26.06 release burndown process.


samson-nvidia (approved):

jakirkham:
@greptile-apps

> All package pins, devcontainer images, and version strings are updated consistently. The only concern is that every shared-workflow reference across build.yaml, pr.yaml, test.yaml, and trigger-breaking-change-alert.yaml was changed from `@cuda-13.2.0` to `@main` — meaning CI now tracks whatever is on the main branch of rapidsai/shared-workflows at the time each run triggers, rather than a known-good snapshot.
>
> .github/workflows/build.yaml, .github/workflows/pr.yaml, .github/workflows/test.yaml, and .github/workflows/trigger-breaking-change-alert.yaml all use `@main` for shared-workflow refs and deserve a second look to confirm a stable 26.08 tag isn't available.

So `rapidsai/shared-workflows` and `rapidsai/shared-actions` are both workflows that live in RAPIDS with the rest of our projects. We intentionally do not pin them as we maintain these workflows and it simplifies making changes to CI across projects

As such we configure zizmor to skip pin checks on these two workflows. Please do the same going forward

Everything else we pin and check they are pinned with zizmor in CI

https://github.com/rapidsai/cugraph-gnn/blob/662a6d9b3be4c4db35f9164e981bbabe8a4e8e54/.github/zizmor.yml#L5-L8

PR #462, opened 2026-05-15
Fp16 embedding train

linhu-nv:
Support BF16/FP16 gradients in embedding backward dedup path


alexbarghi-nv (approved):
Looks good and verified locally that the example runs as expected now.

PR #480, opened 2026-06-25
[DOC] Add Clarifying Comments & Explanations to cuGraph-PyG Examples

alexbarghi-nv:
Adds some clarifying comments and explanations to the cuGraph-PyG examples, including warnings about memory usage and partitioning.


python/cugraph-pyg/cugraph_pyg/examples/mag_lp_mnmg.py
tingyu66, line 615: Can we enforce `local_seeds_per_call=16384,` in test_loader as well? I recall that the default value caused OOM on some SKUs.

tingyu66 (approved):

python/cugraph-pyg/cugraph_pyg/examples/mag_lp_mnmg.py
alexbarghi-nv: Sure, I'll fix that

PR #485, opened 2026-07-03
[FEA] Support Length in Loaders

### LINKED ISSUE #483: [FEA] Support __len__ in Loaders
Support `__len__` in loaders to enable progress tracking.

alexbarghi-nv:
Supports length in loaders, when the user passes a seed list.  Closes #483

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/loader/link_loader.py
+++ b/python/cugraph-pyg/cugraph_pyg/loader/link_loader.py
@@ -1,10 +1,11 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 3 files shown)


python/cugraph-pyg/cugraph_pyg/loader/link_loader.py
```diff
@@ -108,6 +109,7 @@ def __init__(
             )
 
         neg_sampling = torch_geometric.sampler.NegativeSampling.cast(neg_sampling)
+        self.__edge_label_index = edge_label_index
```
tingyu66: This line seems to keep the edge_label_index alive for the loader's lifetime. Would that be expensive considering the loader also creates another copy with clone()? Could we compute if it has explicit labels within __init__, and only store a boolean?

tingyu66 (approved):

python/cugraph-pyg/cugraph_pyg/loader/link_loader.py
alexbarghi-nv: fixed

PR #487, opened 2026-07-08
build CUDA 13 wheels with latest CTK (13.3.0), require C++20 for all builds

### LINKED ISSUE #268: [blocked] wheels: build with newest CUDA 13 CTK
## Description

In RAPIDS 26.04, we switched to building CUDA 13 wheels with CUDA Toolkit 13.0: https://github.com/rapidsai/build-planning/issues/257

This was done to avoid building against libnvJitLink 13.1, which would have brought in a runtime requirement of `nvidia-nvjitlink>=13.1`, which would have made RAPIDS wheels uninstallable against any available `torch` wheels (for CUDA 13).

Eventually we will want to switch back to building RAPIDS wheels against the latest available CTK. This tracks that work.

## Benefits of this work

* take advantage of bugfixes, performance improvements, and new features in new CTKs

## Acceptance Criteria

RAPIDS wheels:

* are built against the latest CTK at the time this is picked up
* all have appropriate runtime dependencies (e.g. `nvidia-nvjitlink` floor updated)
* every project with `torch` wheel tests uses CUDA `torch` wheels for CUDA 12 and 13 *(ideally without needing `pip install --no-deps` to piece together the environment)*

## Approach

* [ ] `ci-imgs` changes
  - [ ] stop building CDUA 13.0.x `ci-wheel` images ()
* [ ] `shared-workflows` changes
* [ ] library changes, in RAPIDS dependency order
  - [ ] rmm
  - [ ] kvikio
  - ~dask-cuda~ *not necessary: pure Python*
  - [ ] ucxx
  - [ ] raft
  - [ ] cuvs
  - [ ] nvforest
  - [ ] cudf
  - [ ] cuopt
  - [ ] cucim
  - [ ] cuxfilter
  - [ ] rapidsmpf
  - [ ] cugraph
  - [ ] cuml
  - [ ] cugraph-gnn
  - ~nx-cugraph~ *not necessary: pure Python*
* [ ] ensure every project is using `@main` / `@release/YY.MM` of `shared-workflows` again

## Notes

N/A

jameslamb:
## Updates to CTK 13 for wheels

Closes https://github.com/rapidsai/build-planning/issues/268

* updates the project's `nvidia-nvjitlink` floor to match that
* expands the `torch` testing matrix... `torch` 2.13 now has a more permissive range of supported `nvidia-nvjitlink` wheels, so **we can now test with `torch` wheels on ALL CUDA 13 versions** 🎉 

Part of a RAPIDS-wide initiative to build wheels using v13.3.0 of the CUDA Toolkit, and to generally get RAPIDS back to the pattern of always building against its latest supported CTK.

As of https://github.com/rapidsai/shared-workflows/pull/590, RAPIDS builds CUDA 13 wheels with v13.3 of the CTK. 

## Updates to C++20

https://github.com/NVIDIA/raft/pull/3052 uses C++20-only `std::atomic<shared_ptr>`. Wholegraph must now also compile with `-std=c++20`, since it has a build-time dependency on RAFT.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,7 +119,7 @@ jobs:
       pull-requests: read
   wheel-build-cugraph-pyg:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@wheels-latest-ctk
     with:
       # This selects "ARCH=amd64 + the latest supported Python + CUDA".
       matrix_filter: map(select(.ARCH == "amd64")) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
@@ -160,7 +160,7 @@ jobs:
       pull-requests: read
   wheel-build-libwholegraph:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@wheels-latest-ctk
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -181,7 +181,7 @@ jobs:
   wheel-build-pylibwholegraph:
     needs: wheel-build-libwholegraph
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@wheels-latest-ctk
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,7 +119,7 @@ jobs:
       pull-requests: read
   wheel-build-cugraph-pyg:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@newest-ctk
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
     with:
       # This selects "ARCH=amd64 + the latest supported Python + CUDA".
       matrix_filter: map(select(.ARCH == "amd64")) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
@@ -160,7 +160,7 @@ jobs:
       pull-requests: read
   wheel-build-libwholegraph:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@newest-ctk
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -181,7 +181,7 @@ jobs:
   wheel-build-pylibwholegraph:
     needs: wheel-build-libwholegraph
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@newest-ctk
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,7 +119,7 @@ jobs:
       pull-requests: read
   wheel-build-cugraph-pyg:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@wheels-latest-ctk
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@newest-ctk
     with:
       # This selects "ARCH=amd64 + the latest supported Python + CUDA".
       matrix_filter: map(select(.ARCH == "amd64")) | group_by(.CUDA_VER|split(".")|map(tonumber)|.[0]) | map(max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]))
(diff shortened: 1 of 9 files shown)


dependencies.yaml
```diff
@@ -500,8 +500,16 @@ dependencies:
               cuda: "13.0"
               require_gpu: "true"
             packages:
-              - *torch_index_cu13
-              - torch==2.10.0+cu130
+              # TODO: switch from '/test/' index (release candidates) once 2.13.0 is officially released
+              - --extra-index-url https://download.pytorch.org/whl/test/cu130
+              - torch==2.13.0+cu130
```
jameslamb: `torch` 2.13 release is planned for tomorrow, July 8, according to https://github.com/pytorch/pytorch/blob/main/RELEASE.md#release-cadence

If that doesn't happen and this ends up getting merged first, I'll make an issue in this repo to track the work of updating to releases.

jameslamb:
Builds (conda and wheels) are failing like this:

```text
CMakeFiles/wholegraph.dir/src/wholegraph_ops/weighted_sample_without_replacement_impl_mapped.cu.o
 │ │ $BUILD_PREFIX/lib/gcc/x86_64-conda-linux-gnu/14.3.0/include/c++/atomic(218): error: static assertion failed with "std::atomic requires a trivially copyable type"
 │ │         static_assert(__is_trivially_copyable(_Tp),
 │ │         ^
 │ │           detected during instantiation of class "std::atomic<_Tp> [with _Tp=std::shared_ptr<raft::resource::resource_factory>]" at line 95 of $PREFIX/include/raft/core/resource/resource_types.hpp
 │ │ $BUILD_PREFIX/lib/gcc/x86_64-conda-linux-gnu/14.3.0/include/c++/atomic(218): error: static assertion failed with "std::atomic requires a trivially copyable type"
 │ │         static_assert(__is_trivially_copyable(_Tp),
 │ │         ^
 │ │           detected during instantiation of class "std::atomic<_Tp> [with _Tp=std::shared_ptr<raft::resource::resource>]" at line 96 of $PREFIX/include/raft/core/resource/resource_types.hpp
 │ │ 2 errors detected in the compilation of "$SRC_DIR/cpp/src/wholegraph_ops/weighted_sample_without_replacement_impl_mapped.cu".
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/28961499832/job/85934108370?pr=487))

Believe this was broken by https://github.com/NVIDIA/raft/pull/3052

We'll need new `cugraph` packages published to fix that.

**update:** merged a small PR to `cugraph` (https://github.com/rapidsai/cugraph/pull/5577), which triggered a new build ([link](https://github.com/rapidsai/cugraph/actions/runs/28965021218)), once that completes this issue should be resolved.

dependencies.yaml
jameslamb: Never mind, 2.13.0 is out, reverted this and now we can use releases 😁

.github/workflows/pr.yaml
```diff
@@ -23,8 +23,10 @@ jobs:
       - wheel-build-libwholegraph
       - wheel-build-pylibwholegraph
       - wheel-tests-pylibwholegraph
+      - wheel-tests-pylibwholegraph-nightly
       - wheel-build-cugraph-pyg
       - wheel-tests-cugraph-pyg
+      - wheel-tests-cugraph-pyg-nightly
```
jameslamb: TODO before merging: remove these

Just added to test that the nightly matrices would work too.

.github/workflows/build.yaml
```diff
@@ -119,7 +119,7 @@ jobs:
       pull-requests: read
   wheel-build-cugraph-pyg:
     secrets: inherit # zizmor: ignore[secrets-inherit]
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@newest-ctk
```
jameslamb: Just picking a random place on the diff so this conversation can be threaded and eventually resolved... still seeing the issues from https://github.com/rapidsai/cugraph-gnn/pull/487#issuecomment-4917476160, so maybe the entire chain of packages picking up new builds for the RAFT ABI break didn't make it up yet.

Just pushed another commit, trying again. If that fails, I'll go look more closely.

jameslamb: A fresh build a few minutes ago broke the same way: https://github.com/rapidsai/cugraph-gnn/actions/runs/29100753622/job/86388952096?pr=487

I think maybe there were timing problems here and we still need some chain of "newest of everything" packages built to unblock CI here. I'll be merging new RAFT, cuVS, and cuGraph PRs as part of https://github.com/rapidsai/build-planning/issues/268 shortly so hopefully that'll do it.

jameslamb: There now new `raft`, `cuvs`, and `cugraph` packages (built in sequence), since https://github.com/rapidsai/cugraph/pull/5575 was merged and built successfully on `main`.

I just pushed a commit here, let's see if CI gets further.

dependencies.yaml
```diff
@@ -268,26 +250,31 @@ dependencies:
...
           - matrix:
-              cuda: "13.2"
+              cuda: "13.0"
               use_cuda_wheels: "true"
             packages:
-              - cuda-toolkit==13.2.*
+              - cuda-toolkit==13.0.*
+          # 'torch' uses '==' pins for its cuda-toolkit dependency and only supports certain versions
+          #
+          # ... allowing this to float allows us
```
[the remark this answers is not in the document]
jameslamb: done in https://github.com/rapidsai/cugraph-gnn/pull/487/commits/abb229194324c53018dfacc759970f125960fb8c, go away.

.github/workflows/build.yaml
jameslamb: Failed the same way, but I think I see the issue... https://github.com/NVIDIA/raft/pull/3052 requires that wholegraph update to C++20. From that PR's description:

> _`std::atomic<shared_ptr>` requires enforcing C++20 for all dependent libraries._

dependencies.yaml
```diff
@@ -268,26 +250,33 @@ dependencies:
...
-              use_cuda_wheels: "true"
-            packages:
-              - cuda-toolkit==13.0.*
-          - matrix:
-              cuda: "13.1"
+              cuda: "12.*"
               use_cuda_wheels: "true"
             packages:
-              - cuda-toolkit==13.1.*
+              - cuda-toolkit>=12.2,<12.8
```
jameslamb: Yes this is weird and non-standard for RAPIDS, but think of it like this... these libraries are complements to / extensions of PyTorch.

If PyTorch chooses not to support CTK 12.2 in wheels, then there's not a real-world use case (other than building both PyTorch and `cugraph-gnn` packages from source) where a user would be able to `pip install torch cuda-toolkit==12.2.* cugraph-pyg-cu12` anyway.

I think this flexibility in CI is justified, in exchange for removing the awkward thing there testing scripts hold knowledge about which CUDA versions we should try to install `torch` for.

alexbarghi-nv (approved):
Should be good to go with the CUDA 12.2 skip

bdice (approved):
Please revert the temporary testing changes. Otherwise LGTM.

dependencies.yaml
jameslamb: I still believe all of this, but ended up reverting it out because CUDA 12.2 CI jobs weren't quite working anyway... in the most recent commits I removed all CUDA 12 related changes and just limited this to expanding testing on CUDA 13.

jameslamb:
Saw all PR CI + nightly CI pass ([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/29125444564/job/86472260474?pr=487)), just pushed https://github.com/rapidsai/cugraph-gnn/pull/487/commits/25861daedd965a96775706a8e66ff257a0de4a9d reverting the nightly matrices out of this.

Also spot-checked some logs to be sure `torch` was getting installed when we inspected it, and with the correct version.

Will queue it for automerge, and then nightly builds should be fixed.

Thanks all!

PR #488, opened 2026-07-10
Pin CMake to < 4.4

KyleFromNVIDIA:
## Description

rapids-cmake is currently broken by CMake 4.4, so pin CMake until we have a fix.

Issue: https://github.com/rapidsai/build-planning/issues/302


gforsyth (approved):
Approving these changes (I was skeptical, but the flowchart convinced me).

CI is broken on `main`, `cmake` isn't the culprit here (but good to get this in)

PR #491, opened 2026-07-15
[FEA] Support Finalizing Graph Store, Improve Edge Index Concatenation

alexbarghi-nv:
Supports finalizing the graph store, removing unneeded tensors from WholeGraph and preventing future modifications of the `GraphStore` object.

Also improves the edge index concatenation process, building the tensor first, then adding to it, then computing the sum, and removing the temporary tensor if the graph store is being finalized.


python/cugraph-pyg/cugraph_pyg/examples/fraud/bitcoin_mnmg.py
```diff
@@ -102,6 +102,7 @@ def init_pytorch_worker(global_rank, local_rank, world_size, cugraph_id):
     ] = data.edge_index if rank == 0 else empty(dim=2)
     feature_store["entity", "x", None] = data.x if rank == 0 else empty(dim=2)
     feature_store["entity", "y", None] = data.y if rank == 0 else empty(dim=1)
+    graph_store.finalize
```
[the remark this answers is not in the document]
alexbarghi-nv: fixed

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -246,6 +301,10 @@ def _graph(self) -> Union[pylibcugraph.SGGraph, pylibcugraph.MGGraph]:
 
         return self.__graph
```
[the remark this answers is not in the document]
alexbarghi-nv: fixed

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -30,6 +30,20 @@
...
+        raise ValueError(
+            "This GraphStore object has been finalized,"
+            " and the edge index is no longer available."
+        )
+
+    def __getitem__(self, key):
+        raise ValueError(
+            "This GraphStore object has been finalized,"
+            " and the edge index is no longer available."
+        )
```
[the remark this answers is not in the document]
alexbarghi-nv: fixed

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -243,9 +298,15 @@ def _graph(self) -> Union[pylibcugraph.SGGraph, pylibcugraph.MGGraph]:
                     if "etime" in edgelist_dict
                     else None,
                 )
```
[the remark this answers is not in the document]
alexbarghi-nv: fixed

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -192,23 +243,27 @@ def _resource_handle(self):
...
+    ) -> Union[pylibcugraph.SGGraph, pylibcugraph.MGGraph]:
         graph_properties = pylibcugraph.GraphProperties(
             is_multigraph=True, is_symmetric=False
         )
 
         if self.__graph is None:
-            edgelist_dict = self.__get_edgelist()
+            edgelist_dict = self.__get_edgelist(finalize=finalize)
 
+            num_vertices = sum(self._num_vertices().values())
```
tingyu66, line 266: When `finalize=True`, __get_edgelist() replaces the stored indices with EmptyEdgeIndex. The following _num_vertices() would fail if EdgeAttr.size is None. Would EdgeAttr.size always be provided?

alexbarghi-nv: The size is typically provided on creation, and I think it's acceptable for it to fail telling the user that they can't calculate the number of vertices now that the edge index object is no longer available.  `_num_vertices()` is also a semi-private method so it's not really meant to be called unless you understand the implementation details.

Also this code won't be called after `finalize()` - the finalization instantiates the SGGraph or MGGraph so `__graph` won't be `None` here.

alexbarghi-nv: Oh actually I see what you're pointing out here - it fails during finalization if the sizes weren't provided, not after.  Ok, let me refactor this.

tingyu66: Thanks for the fix. Looks good to me now. I’ll leave it up to you whether to add a regression test that calls put_edge_index(..., size=None) before finalize() to improve coverage when sizes are absent.

tingyu66 (approved):

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
alexbarghi-nv: In the next PR I may add some more testing that covers this and other things.

PR #497, opened 2026-07-21
Support Loading Parquet Files in WholeGraph

alexbarghi-nv:
## Summary

Adds direct Parquet input support for WholeMemory tensors and embeddings, including the cuGraph-PyG `DistTensor` and `DistEmbedding` file-loading APIs.

The implementation is designed for datasets that may be several terabytes: it allocates the final WholeMemory destination first, then decodes and copies only bounded Parquet batches into each rank's local partition. It never loads or concatenates the complete source dataset in host memory.

## Implementation

- supports `binary`, `parquet`, and `auto` file formats
- uses Parquet footer metadata to infer and optionally validate tensor shape before allocation
- makes `expected_shape` optional for file input; it is used only for validation and one-column dimensionality disambiguation
- infers `last_dim_size` for Parquet input and requires it for raw binary input
- treats one-column Parquet input as a 1-D `(N,)` tensor by default; `expected_shape=(N, 1)` requests the 2-D representation
- keeps the destination `dtype` explicit and warns when Parquet column types differ before converting bounded batches
- adds `fail_on_dtype_mismatch=True` to turn that warning into a metadata-validation error before allocation
- treats a file list as one logical row-major tensor
- maps each rank's contiguous WholeMemory row interval onto the input files and reads only intersecting row groups and rows
- converts one bounded columnar batch at a time into the dense row-major layout required by WholeMemory
- writes batches directly into the final CPU or CUDA WholeMemory allocation
- preserves the native binary loader as the lowest-overhead path
- supports explicit `tensor_entry_partition` / `partition_book` values; otherwise WholeMemory applies its default partitioning
- keeps round-robin loading binary-only

## Why direct PyTorch and NumPy file input are not supported

PyTorch's public file-loading API does not provide a bounded, windowed read interface. `torch.load(..., mmap=True)` avoids eagerly copying the complete tensor, but pages touched during a sequential scan can remain resident while the full-file mapping is open. Keeping RSS strictly bounded would therefore depend on platform-specific page-eviction calls, while older PyTorch serialization formats may not support mmap at all.

That behavior is not a good fit for a loader whose primary requirement is predictable host-memory use at multi-terabyte scale. Callers should use Parquet for structured input or convert data to the native binary format for the lowest possible overhead. `.pt` and `.pth` files are explicitly rejected when format detection is used.

The previous `.npy` path had the same issue: `np.load(..., mmap_mode="c")` avoided an eager copy, but pages touched while copying a rank's partition could remain resident and make peak RSS scale with that partition. Direct `.npy` loading has therefore also been removed. Known `.pt`, `.pth`, and `.npy` extensions are rejected even when callers attempt to force the binary format, preventing serialized headers from being misinterpreted as raw tensor data.

## Memory behavior and tests

The tests cover:

- Parquet correctness for CPU and CUDA WholeMemory destinations
- Parquet integration through cuGraph-PyG `DistTensor` and `DistEmbedding`
- metadata-only row-count and schema validation
- inferred multi-column shape and optional expected-shape validation
- one-column `(N,)` and `(N, 1)` behavior
- warnings for Parquet/destination dtype mismatches
- strict dtype mismatch rejection through the low-level and cuGraph-PyG APIs
- explicit rejection of direct `.pt`, `.pth`, and `.npy` input
- requested row ranges and multi-row-group input
- non-numeric column and shape validation
- end-to-end peak host RSS for CPU and CUDA destinations
- reader-only peak host RSS
- memory scaling from 32 MiB to 128 MiB datasets in fresh subprocesses

The scaling tests subtract only the required CPU WholeMemory destination allocation and verify that the remaining reader/conversion overhead stays within a fixed bound and does not grow materially when the dataset size increases 4x.

## Local validation

- pylibwholegraph Parquet suite: `21 passed`
- cuGraph-PyG Parquet `from_file` integration: `4 passed`
- pre-commit hooks: passed
- `git diff --check`: passed


linhu-nv (approved):
This looks good to me.

tingyu66 (approved):

dependencies.yaml
```diff
@@ -395,6 +396,7 @@ dependencies:
     common:
       - output_types: [conda, pyproject, requirements]
         packages:
+          - *pyarrow
```
jameslamb, line 429:
```suggestion
```

Since `pyarrow` is now a hard required dependency, I'd keep it out of the `[test]` extra. I think we generally prefer to only have that extra provide additional packages on top of what's in `dependencies` or to provide different constraints.

For example, `cudf` declares `pyarrow` in its runtime dependencies but not its `[test]` extra: https://github.com/rapidsai/cudf/blob/cfb5a7d8ce9c902844e216054a49b36daebc5491/python/cudf/pyproject.toml#L34

python/pylibwholegraph/pylibwholegraph/torch/tensor.py
```diff
@@ -16,11 +19,356 @@
 from .wholegraph_env import wrap_torch_tensor, get_wholegraph_env_fns, get_stream
 
 torch = import_optional("torch")
+np = import_optional("numpy")
+pa = import_optional("pyarrow")
+pq = import_optional("pyarrow.parquet")
```
jameslamb, line 24: `numpy` and `pyarrow` are hard required dependencies for `pylibwholegraph`... what's the benefit of passing them through the `import_optional()` machinery here?

I think either:

* they should just be unconditional imports
* `pyarrow` should move to an optional dependency via an extra like `pylibwholegraph[pyarrow]`

Right now this is paying the complexity cost of optional imports without the compensating benefit of the dependencies being lighter-weight.

python/pylibwholegraph/pylibwholegraph/torch/tensor.py
```diff
@@ -16,11 +19,356 @@
...
+    """Map supported scalar Arrow types without going through pandas."""
+    if pa.types.is_boolean(parquet_type):
+        return np.dtype(np.bool_)
+    if pa.types.is_signed_integer(parquet_type):
+        return np.dtype(f"int{parquet_type.bit_width}")
+    if pa.types.is_unsigned_integer(parquet_type):
+        return np.dtype(f"uint{parquet_type.bit_width}")
+    if pa.types.is_floating(parquet_type):
+        return np.dtype(f"float{parquet_type.bit_width}")
+    raise ValueError(f"unsupported Parquet column type {parquet_type}")
```
jameslamb:
```suggestion
    raise ValueError(f"unsupported Parquet column type '{parquet_type}'")
```

I'd put this type in single quotes, just in case the input here ever looks like an English word. For example, imagine you encountered this error:

> ValueError: unsupported Parquet column type string

You might think "what's a column type string?".

I think that's less likely if it looks like this:

> ValueError: unsupported Parquet column type 'string'

python/pylibwholegraph/pyproject.toml
```diff
@@ -29,13 +29,15 @@ classifiers = [
 dependencies = [
     "libwholegraph==26.10.*,>=0.0.0a0",
     "numpy>=2.0,<3.0",
+    "pyarrow",
```
jameslamb, line 32: Just to write this somewhere... I think it's totally fine to start with this unconstrained and only add constraints as you discover issues, especially with the small slice of the `pyarrow` API this PR is using.

jameslamb (approved):
Reviewed for `packaging-codeowners`, left some very very minor feedback. Do what you want with it, including ignoring it or addressing in a follow-up. Marking this "approve" so you have what you need to merge without another review.

I only lightly skimmed the Python code, deferring to the other reviews on correctness, performance, etc.

python/pylibwholegraph/pylibwholegraph/torch/tensor.py
alexbarghi-nv: fair; I made the change

python/pylibwholegraph/pyproject.toml
alexbarghi-nv: Actually now that I think about it - do we have a RAPIDS-wide pyarrow constraint?

alexbarghi-nv: Not that we should apply it here but just as something that's nice to know

jameslamb: Good question. I think the answer is "not really, everything just kind of follows what `pandas` / cuDF do".

Looks like `cudf` today is the only one with an explicit pin on it: https://github.com/search?q=pyarrow+path%3Adependencies.yaml+%28org%3Arapidsai+OR+org%3ANVIDIA%29+AND+NOT+is%3Aarchived&type=code

PR #503, opened 2026-07-23
Drop unused Python test dependencies

alexbarghi-nv:
## Summary

- remove unused `torchdata` and `pydantic` entries from the Python test dependency matrix
- regenerate the tracked conda development and all-dependencies environments

## Impact

cuGraph-GNN test environments no longer install the unused `torchdata` and `pydantic` packages.

## Validation

- `pre-commit run rapids-dependency-file-generator --all-files`
- `git diff --check`
- confirmed no remaining `torchdata` or `pydantic` references outside changelog/history

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -29,7 +29,6 @@ dependencies:
 - numpy>=2.0,<3.0
 - pandas
 - pre-commit
-- pydantic
 - pylibcugraph==26.10.*,>=0.0.0a0
 - pytest
 - pytest-benchmark
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -42,6 +42,5 @@ dependencies:
 - scikit-build-core>=0.11.0
 - scipy
 - setuptools>=77.0.0
-- torchdata
 - wheel
 name: all_cuda-129_arch-aarch64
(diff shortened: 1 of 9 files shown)


jameslamb (approved):
I vaguely recall looking into these before and finding a comment in code (here? cuGraph? don't remember) that these were needed to provide optional dependencies for some relevant code path in `sentence-transformers` or `ogb` or something.

Anyway... can't find that and if you're happy with what you see in CI (nothing getting skipped that shouldn't, etc.) then great, happy to see these drop out of the testing dependencies!

alexbarghi-nv:
> I vaguely recall looking into these before and finding a comment in code (here? cuGraph? don't remember) that these were needed to provide optional dependencies for some relevant code path in `sentence-transformers` or `ogb` or something.
> 
> Anyway... can't find that and if you're happy with what you see in CI (nothing getting skipped that shouldn't, etc.) then great, happy to see these drop out of the testing dependencies!

`torchdata` was for DGL I believe.
`pydantic` I'm not sure about, but it's definitely not needed for CI. Let me double-check that none of the examples need it.

alexbarghi-nv:
> > I vaguely recall looking into these before and finding a comment in code (here? cuGraph? don't remember) that these were needed to provide optional dependencies for some relevant code path in `sentence-transformers` or `ogb` or something.
> > Anyway... can't find that and if you're happy with what you see in CI (nothing getting skipped that shouldn't, etc.) then great, happy to see these drop out of the testing dependencies!
> 
> `torchdata` was for DGL I believe. `pydantic` I'm not sure about, but it's definitely not needed for CI. Let me double-check that none of the examples need it.

I've confirmed again that pydantic can safely be dropped.  I think it was also for DGL.

jameslamb:
ahhhh DGL, ok. Thanks for double-checking that.

akihironitta (approved):

PR #505, opened 2026-07-23
Add opt-in RMM allocation for WholeMemory

alexbarghi-nv:
## Summary

- add C API controls and `pylibwholegraph.torch.set_memory_resource()` / `is_rmm_enabled()` to opt supported WholeMemory allocations into RMM
- allocate distributed and hierarchy device storage with CCCL's `cuda::device_buffer<cuda::std::byte>`, explicitly backed by RMM's current per-device resource
- preserve byte-based sizing, uninitialized storage, CUDA allocation alignment, and legacy default-stream behavior
- retain the allocation-time resource reference for matching deallocation; the Python API also retains supplied memory-resource objects for allocation lifetime safety
- keep chunked, continuous, and NVSHMEM device storage on their specialized allocation paths and emit a rank-0 fallback warning when RMM is enabled
- leave host allocations and the default disabled path unchanged, require matching enablement across ranks, and reset RMM enablement during successful WholeMemory finalization
- add RMM runtime dependencies to conda and Python package metadata, plus Python usage documentation

## Test coverage

- verify the default/disabled path does not allocate through RMM
- verify distributed and hierarchy device allocations use the configured RMM resource and release all tracked bytes
- verify an allocation is deallocated through its original resource after RMM's current resource is changed
- verify chunked and continuous device allocations, plus distributed host allocations, remain outside RMM
- verify finalization resets RMM enablement
- verify the Python API installs and retains a valid resource, supports disabling with `None`, rejects invalid resource types, and reports finalization state correctly

## Local testing

Tested locally in the `rapids` conda environment with CUDA 12.9, GCC 14, RMM 26.10, CCCL 3.4, and NVIDIA V100 GPUs:

- fresh CMake configure and CUDA build of `WHOLEMEMORY_HANDLE_TEST` (passed)
- complete `WHOLEMEMORY_HANDLE_TEST` on two peer-connected V100s: **20 tests passed**
- focused `WholeMemoryHandleRMMTests.*` two-GPU run (passed)
- new Python memory-resource API tests: **3 tests passed**
- existing Python WholeMemory binding regression: **1 test passed**
- repository pre-commit hooks for all changed files, including clang-format, Ruff, RAPIDS dependency generation, CMake formatting/linting, copyright checks, and alpha-spec validation (passed)
- `git diff --check` (passed)


cpp/src/wholememory/memory_handle.cpp
```diff
@@ -363,7 +376,11 @@ class distributed_wholememory_impl : public wholememory_impl {
       return;
     }
 
-    if (on_device) {
+    if (on_device && is_rmm_enabled()) {
+      rmm_device_buffer_ =
+        std::make_unique<rmm::device_buffer>(alloc_size, rmm::cuda_stream_default);
```
bdice: We are going to be migrating all libraries from `rmm::device_buffer` to the `cuda::buffer` class from CCCL. If your only goal is to use the current device resource from RMM, try using `cuda::buffer` and pass the RMM resource to its constructor. See https://github.com/rapidsai/rmm/issues/2455 for more details.

bdice (commented):
Generally looks fine, but I have a comment about how this code could be a little more future-proof.

cpp/src/wholememory/memory_handle.cpp
alexbarghi-nv: Oh wow, you were fast!  I hadn't even intended for this to get reviewed yet, it's still in draft

alexbarghi-nv: I will incorporate your suggestion

alexbarghi-nv: Thanks, this makes sense. I’m planning to keep RMM as the source of the current per-device resource, but use CCCL’s owning buffer type as follows:

```diff
-#include <rmm/device_buffer.hpp>
+#include <cuda/buffer>
+#include <cuda/memory_resource>
+#include <cuda/std/cstddef>
+#include <rmm/mr/per_device_resource.hpp>

@@
     if (on_device && is_rmm_enabled()) {
+      auto const alignment_env = cuda::std::execution::prop{
+        cuda::allocation_alignment,
+        cuda::mr::default_cuda_malloc_alignment};
+
       rmm_device_buffer_ =
-        std::make_unique<rmm::device_buffer>(alloc_size, rmm::cuda_stream_default);
+        std::make_unique<cuda::device_buffer<cuda::std::byte>>(
+          cuda::stream_ref{rmm::cuda_stream_default.value()},
+          rmm::mr::get_current_device_resource_ref(),
+          alloc_size,
+          cuda::no_init,
+          alignment_env);
       dev_ptr = rmm_device_buffer_->data();
@@
-  std::unique_ptr<rmm::device_buffer> rmm_device_buffer_;
+  std::unique_ptr<cuda::device_buffer<cuda::std::byte>> rmm_device_buffer_;
```

Using a byte element type keeps the size byte-based, `cuda::no_init` avoids unnecessary initialization, and the alignment environment preserves the existing 256-byte CUDA allocation alignment. The current RMM resource is passed explicitly and remains the resource used when the buffer is destroyed. I’ll keep the existing tests and RMM dependencies unchanged.

bdice: Replace `cuda::stream_ref{rmm::cuda_stream_default.value()}` with `cuda::stream_ref{cudaStream_t{nullptr}}` and I think it's about right.

bdice: Also maybe rename it to `device_buffer_` rather than `rmm_device_buffer_`.

alexbarghi-nv: Done.

python/pylibwholegraph/pylibwholegraph/torch/initialize.py
```diff
@@ -13,6 +13,50 @@
...
+        raise ImportError(
+            "Setting a WholeMemory memory resource requires the Python rmm package."
+        ) from exc
+
+    if not isinstance(memory_resource, rmm.mr.DeviceMemoryResource):
+        raise TypeError("memory_resource must be an rmm.mr.DeviceMemoryResource")
+
+    rmm.mr.set_current_device_resource(memory_resource)
+    if all(memory_resource is not resource for resource in _memory_resources):
+        _memory_resources.append(memory_resource)
```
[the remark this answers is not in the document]
alexbarghi-nv: resolved (by removing this API)

python/pylibwholegraph/pylibwholegraph/torch/initialize.py
```diff
@@ -15,6 +15,25 @@
...
+    When enabled, future distributed and hierarchy WholeMemory tensors with device storage use
+    RMM's current resource for the active CUDA device at allocation time. Configure that resource
+    through RMM before creating the tensors. Chunked, continuous, and NVSHMEM device tensors retain
+    their specialized allocation paths.
+
+    :param enabled: Whether supported WholeMemory device allocations should use RMM.
+    :return: None
+    """
+    wmb.set_rmm_enabled(enabled)
+
```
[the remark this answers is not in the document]
alexbarghi-nv: This comment is wrong.

cuda::device_buffer does retain the allocation-time resource. Although its constructor receives an RMM resource_ref, the buffer converts that into an owned cuda::mr::any_resource and stores it alongside the allocation. Current RMM pool resources use shared ownership, so replacing the Python current resource and releasing its Python wrapper does not invalidate the copy held by device_buffer_. 

Codex also verified this for me by creating a test that replaced the current resource, deleted the original Python resource, forced GC, then successfully destroyed the WholeMemory allocation.

cpp/CMakeLists.txt
```diff
@@ -107,6 +107,7 @@ rapids_cpm_init()
 
 include(./cmake/thirdparty/get_raft.cmake)
 include(./cmake/thirdparty/get_nccl.cmake)
+find_package(rmm "${RAPIDS_VERSION}" REQUIRED)
```
bdice: Please copy https://github.com/rapidsai/cudf/blob/main/cpp/cmake/thirdparty/get_rmm.cmake, don't use plain `find_package` here. This must go _before_ `get_raft.cmake` so that RMM is found by wholegraph (potentially with custom branches/etc.) _before_ RAFT attempts to find RMM.

tingyu66 (approved):

python/pylibwholegraph/README.md
```diff
@@ -1,5 +1,31 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
bdice, line 2: If you want to remove this, I think README files are exempt from the requirement of having copyright headers. I don't see it being checked here: https://github.com/rapidsai/cugraph-gnn/blob/674398fd39bdf95ec6c7750dbe56bbcb8304c9e0/.pre-commit-config.yaml#L45-L59

Up to you.

bdice (changes requested):
One CMake issue needs to be fixed before we can merge.

cpp/CMakeLists.txt
alexbarghi-nv: done; can you re-review @bdice ?

python/pylibwholegraph/README.md
alexbarghi-nv: It's not a huge deal; I'll leave it in

linhu-nv (approved):
LGTM thanks!

cpp/src/wholememory/memory_handle.cpp
```diff
@@ -363,7 +379,17 @@ class distributed_wholememory_impl : public wholememory_impl {
       return;
     }
 
-    if (on_device) {
+    if (on_device && is_rmm_enabled()) {
+      auto const alignment_env = cuda::std::execution::prop{
+        cuda::allocation_alignment, cuda::mr::default_cuda_malloc_alignment};
+      device_buffer_ = std::make_unique<cuda::device_buffer<cuda::std::byte>>(
+        cuda::stream_ref{cudaStream_t{nullptr}},
+        rmm::mr::get_current_device_resource_ref(),
```
linhu-nv, line 387: my agent told me this should be replaced with a owning ref using shared_ptr<rmm::mr::device_memory_source> instead of a non-owning one like now, otherwise there may be crash if the rmm pool is freed before wholememory handle.

alexbarghi-nv: I agree with your agent, but I'm not sure we can just create a shared pointer here.

`cuda::device_buffer` owns a copy of the resource object passed to it, but we pass:

```cpp
rmm::mr::get_current_device_resource_ref()
```

That is a non-owning `device_async_resource_ref`. Copying it into the buffer only copies the reference; it does not extend the underlying resource’s lifetime.

Therefore:

- Dereferencing the user’s Python/C++ variable is safe while RMM still owns that resource as the current per-device resource.
- Replacing it through `rmm.reinitialize()` while a WholeMemory allocation is alive can destroy the final owner.
- The buffer would then retain a dangling resource reference and invoke undefined behavior when deallocating.

Our C++ test doesn’t catch this because both `statistics_mr` objects remain alive on the stack until after the buffers are destroyed.

The complication is that we cannot safely construct shared ownership from an already non-owning `device_async_resource_ref`. Wrapping that reference in a shared object would only keep the reference alive, not its target. We would need either:

- an RMM API returning an owning `cuda::mr::any_resource`/shared resource for the current device, or
- to document and enforce that users must destroy WholeMemory allocations before calling `rmm.reinitialize()`.

So yes: the buffer should ideally retain shared ownership, but the current RMM getter does not give us enough ownership information to do that generically. The current implementation is only safe under the usual RMM resource-lifetime contract.

I'm going to push an update that clarifies that all WM allocations should be destroyed before changing the current RMM device memory resource.

alexbarghi-nv: Done.

bdice (approved):
Nice! Thank you.

PR #508, opened 2026-07-28
ensure nightly builds always produce new packages

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/218

* uses the new patterns for versioning nightly packages (see https://github.com/rapidsai/shared-workflows/pull/603), to ensure they're always uploaded on new builds. 

For wheels:

```text
# before
26.10.0a55 

# after
26.10.0a55.post260803200854
```

And for conda:

```text
# before
26.10.0a55[build=cuda12_260803_9da146ef]

# after
26.10.0a55[build=cuda12_260803200854_9da146ef]
```

Other changes:

* updates `changed-files` lists to avoid triggering test jobs in PR CI when only `.github/workflows/{build,test}.yaml` are changed

### How I tested this

<details><summary>round 1: epoch seconds versions (click me)</summary>

Triggered a build from this branch.

```shell
gh workflow run \
  build.yaml \
  --ref nightly-versions \
  --repo rapidsai/cugraph-gnn \
  -f branch=nightly-versions \
  -f build_type=branch \
  -f date=2026-07-27 \
  -f sha="$(git rev-parse HEAD)"
```

That started this: https://github.com/rapidsai/cugraph-gnn/actions/runs/30321225539

Then did a second run, to show that new packages would be published from the same commit: https://github.com/rapidsai/cugraph-gnn/actions/runs/30322425531

Then triggered a testing run.

```shell
gh workflow run \
  test.yaml \
  --ref nightly-versions \
  --repo rapidsai/cugraph-gnn \
  -f branch=nightly-versions \
  -f build_type=branch \
  -f date=2026-07-27 \
  -f sha="$(git rev-parse HEAD)"
```

Build link: https://github.com/rapidsai/cugraph-gnn/actions/runs/30323823418

That all looks good!

The second build uploaded new packages with the same beginning version (`26.10.0a21`) but different suffixes (`.post1785203206` vs. `.post1785204732`).

All packages from all jobs within a run had the same version (`26.10.00a21.post1785204732`).

The newer packages were successfully pulled by all the test jobs. e.g. output like this in a conda job:

```text
...
  + libwholegraph                     26.10.00a21.post1785204732  cuda12_260728_9fc66978             /tmp/tmp.bcD2HOYsle      49MB
...
  + pylibwholegraph                   26.10.00a21.post1785204732  cuda12_cp311_abi3_260728_9fc66978  /tmp/tmp.GAjTiyLk9l     298kB
...
```

And this in a wheels job:

```text
Successfully installed ... libwholegraph-cu13-26.10.0a21.post1785204732
... pylibwholegraph-cu13-26.10.0a21.post1785204732 ...
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/30323823418/job/90164980602))

</details>

<details><summary>round 2: YYMMDDhhmmss versions (click me)</summary>

During review, we switched to `YYMMDDhhmmss`-formatted versions strings (https://github.com/rapidsai/shared-workflows/pull/603#discussion_r3667669245), so I repeated the testing above with that new version format.

* `build.yaml` 1: https://github.com/rapidsai/cugraph-gnn/actions/runs/30396789538
* `build.yaml` 2: https://github.com/rapidsai/cugraph-gnn/actions/runs/30399476266
* `test.yaml` : https://github.com/rapidsai/cugraph-gnn/actions/runs/30402598280/job/90422013211
  - _just 1 conda job failed, and the failure looks unrelated to these changes_

</details>

After more review, we switched away from updating *versions* for conda packages, towards just replacing the YYMMDD component of *build strings* with a more precise date-time (https://github.com/rapidsai/cugraph-gnn/pull/508#discussion_r3678533912), so repeated the testing again.

* `build.yaml` 1: https://github.com/rapidsai/cugraph-gnn/actions/runs/30648937831
* `build.yaml` 2: https://github.com/rapidsai/cugraph-gnn/actions/runs/30652238487
* `test.yaml`: https://github.com/rapidsai/cugraph-gnn/actions/runs/30653529575


ci/build_cpp.sh
```diff
@@ -1,9 +1,18 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 set -euo pipefail
 
+if ! [ -d /tmp/gha-tools ]; then
```
jameslamb: TODO before merging... remove all of these once these PRs are merged:

* https://github.com/rapidsai/gha-tools/pull/268
* https://github.com/rapidsai/shared-workflows/pull/603

.github/workflows/build.yaml
```diff
@@ -31,24 +31,28 @@ concurrency:
...
     uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@main
+    needs: [build-details]
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       node_type: cpu8
       script: ci/build_cpp.sh
       sha: ${{ inputs.sha }}
+      version-suffix: ${{ needs.build-details.outputs.package-version-suffix }}
```
bdice: I should've realized this earlier, but this suffix only makes sense for wheels (it contains `.post`) and yet its name doesn't indicate anything about wheels.

We aren't going to suffix conda versions, we're just going to use build strings, so I don't want to imply that this is generic to all build types.

We should name this something like `RAPIDS_DATETIME_STRING` (to align with conda's usage of `RAPIDS_DATE_STRING`) and then use `.post${RAPIDS_DATETIME_STRING}` for wheels and insert `${RAPIDS_DATETIME_STRING}` into the conda recipes where we currently use `RAPIDS_DATE_STRING`.

jameslamb: > _We aren't going to suffix conda versions, we're just going to use build strings_

As written, I AM proposing using this suffix conda packages too and that's how it's working right now:

```text
...
cugraph-pyg                         26.10.00a21.post260728210952  260728_9fc66978                    file:///tmp/tmp.vYeKSGpHdu
...
```

([example from the test runs linked in the description](https://github.com/rapidsai/cugraph-gnn/actions/runs/30402598280/job/90422013345#step:12:4240))

Because:

1. I wasn't sure how the build string participates in sorting (would the latest nightly always be preferred?)
2. I didn't want to mess with the existing YYMMDD component in the build strings and break the `test-nightly-conda-env` stuff in `integration` ([code link](https://github.com/rapidsai/integration/blob/3ca79d0fea7ed9339bcee66556186085958ff963/ci/check_conda_nightly_env.py#L36-L37)) and anything else that might be relying on that formatting.

Fine with me to change the naming of these things and try what you're suggesting, and I'll happily patch `integration` and anything else that needs to change as a result of changing the build string.

bdice: I think there may be issues with conda solvers with both pre- and post- suffixes. I am not 100% sure about that. I do think we should stick to the status quo for conda, aside from adopting the same build string date format proposed for wheels.

The only place we explicitly rely on that date format string is in the `integration` repo's parser, to my knowledge.

ci/build_wheel.sh
```diff
@@ -42,14 +42,24 @@ if [[ "${package_name}" != "libwholegraph" ]]; then
...
-source rapids-date-string
+source rapids-datetime-string
 source rapids-init-pip
 
 export SCCACHE_S3_PREPROCESSOR_CACHE_KEY_PREFIX="${package_name}/${RAPIDS_CONDA_ARCH}/cuda${RAPIDS_CUDA_VERSION%%.*}/wheel/preprocessor-cache"
 export SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
 
-rapids-generate-version > ./VERSION
+RAPIDS_VERSION_SUFFIX=".post${RAPIDS_DATETIME_STRING}" \
+  rapids-generate-version > ./VERSION
```
jameslamb, line 53: This hits this code path: https://github.com/rapidsai/gha-tools/blob/02ea00f8c3b257c8aeb94cb757724932e70ce0e2/tools/rapids-generate-version#L26

That's already guarded by `rapids-is-release-build`, so this won't result in a `.post` version being added to releases... just nightlies and PR artifacts.

.github/workflows/pr.yaml
```diff
@@ -10,6 +10,7 @@ permissions: {}
 jobs:
   pr-builder:
     needs:
+      - build-details
```
jameslamb, line 13: In the first attempt at https://github.com/rapidsai/build-planning/issues/218 I'd tried to limit this change to only branch / nightly builds, but now that we're doing a bit more than just "stick an optional suffix on the version number", I think we should just expand this to PRs too.

Reasons:

1. makes flow of configuration stricter and therefore simpler
2. can use PR CI to test changes to this (and detect issues earlier)

.github/workflows/pr.yaml
```diff
@@ -59,6 +62,8 @@ jobs:
           - '!.github/copy-pr-bot.yaml'
           - '!.github/ops-bot.yaml'
           - '!.github/release.yml'
+          - '!.github/workflows/build.yaml'
+          - '!.github/workflows/test.yaml'
```
jameslamb, line 66: Planning to tack this onto all of these PRs, to take advantage of the reviews and to set us up for changes like https://github.com/rapidsai/build-planning/issues/278 being cheaper.

.github/workflows/pr.yaml
```diff
@@ -31,6 +32,8 @@ jobs:
       needs: ${{ toJSON(needs) }}
     permissions:
       contents: read
+  build-details:
```
jameslamb, line 35: (picking arbitrary place just so this can be a thread)

All Python tests are failing like this:

```text
>   ???
E   TypeError: homogeneous_biased_temporal_neighbor_sample() got an unexpected keyword argument 'starting_vertex_times'
```

([build link](https://github.com/rapidsai/cugraph-gnn/actions/runs/30648496959/job/91218343195?pr=508#step:13:2823))

Looks to me like `cugraph-gnn` projects need updates to react to this: https://github.com/rapidsai/cugraph/pull/5603

ci/build_cpp.sh
jameslamb: This is back in the diff so unresolving... these should all be reverted before merging.

bdice (approved):

jameslamb:
Spot-checked some logs for conda and wheels and it looks like this is all working as expected. It's a technicality that @bdice approved while this was still in draft and therefore GitHub still sees this as needing a `packaging-codeowners` approval... I haven't made any substantive changes since making it ready for review.

Admin-merging this (with fully passing CI!).

PR #511, opened 2026-07-28
Fix dist matrix local access

alexbarghi-nv:
## Description

Fixes a race where `DistMatrix` local accessors could return stale or uninitialized data immediately after WholeGraph tensor initialization or updates.

WholeGraph scatter and copy operations run on the current PyTorch CUDA stream, while the WholeGraph communicator barrier uses an internal stream. A barrier alone therefore does not guarantee that preceding PyTorch stream work has completed.

This change explicitly synchronizes the current PyTorch CUDA stream before entering the communicator barrier.

## Changes

- Synchronize the current PyTorch CUDA stream before WholeGraph communicator barriers during:
  - distributed tensor initialization
  - `DistMatrix` updates
- Restore zero-copy `local_col` and `local_row` access using `get_local_tensor()`.
- Remove the distributed gathers previously used to work around the synchronization race.
- Avoid constructing a full-size index tensor for partial slice assignments.
- Add multi-GPU tests covering:
  - visibility of updates through local accessors
  - zero-copy storage sharing
  - non-uniform WholeGraph partitions

## Memory impact

The zero-copy local accessors avoid temporary allocations previously required for:

- CPU index tensors
- copied CUDA index tensors
- CUDA gather outputs

Partial slice assignments also allocate indices only for the selected range instead of first constructing an index tensor for the entire matrix.

## Performance considerations

Updates now include a CUDA stream synchronization and communicator barrier to guarantee visibility across ranks. This adds synchronization overhead to updates, but removes the gather and allocation overhead from every local accessor call.

## Testing

```text
python -m pytest python/cugraph-pyg/cugraph_pyg/tests/tensor -q

19 passed

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/tensor/dist_matrix.py
+++ b/python/cugraph-pyg/cugraph_pyg/tensor/dist_matrix.py
@@ -5,6 +5,7 @@ from typing import Optional, Union, Tuple, List, Literal
 
 from cugraph_pyg.utils.imports import import_optional
 from cugraph_pyg.tensor import DistTensor
+from cugraph_pyg.tensor.utils import synchronize_stream_and_barrier
 
 torch = import_optional("torch")
 
@@ -94,7 +95,8 @@ class DistMatrix:
     ):
         if isinstance(idx, slice):
             size = self._col.shape[0]
-            idx = torch.arange(size)[idx]
+            start, stop, step = idx.indices(size)
+            idx = torch.arange(start, stop, step)
 
         if self._format != "coo":
             raise ValueError("Updating is currently only supported for COO format")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by alexbarghi-nv** [comment]
Forgot to rebase, sorry

File: python/cugraph-pyg/cugraph_pyg/tensor/dist_matrix.py [hunk 1]
```diff
@@ -177,35 +185,11 @@ def transpose(self, view: bool = False) -> "DistMatrix":
...
-
-        q = sz // world_size
-        r = sz % world_size
-
-        if rank < r:
-            ix = torch.arange(q * rank + rank, q * (rank + 1) + rank + 1)
-        else:
-            ix = torch.arange(q * rank + r, q * (rank + 1) + r)
-        return self._row[ix]
+        return self._row.get_local_tensor()
```
tingyu66, line 196: Just a nitpick here. local_col and local_row now expose tensor’s actual partition, so local_coo would only work when col and row have the same partition_book. Maybe we should enforce this behavior for COO format in the constructor?

tingyu66 (approved):

python/cugraph-pyg/cugraph_pyg/tensor/dist_matrix.py
alexbarghi-nv: I fixed the constructor; this should be resolved now

PR #527, opened 2026-09-01
Add last temporal neighbor sampling to cuGraph-PyG

alexbarghi-nv:
## Summary

- map `temporal_strategy=last` to the unified pylibcugraph `neighbor_selection=last` API
- update the fixed-window integration from #524 for the orthogonal `fixed_window` flag introduced by rapidsai/cugraph#5636
- require both lower and upper seed-time bounds for fixed-window last sampling
- reject replacement and biased sampling combinations that libcugraph does not support
- warn that integer edge timestamps above `2**53` may be ranked inexactly
- add homogeneous multi-hop coverage for selecting the newest eligible edge

## Testing

- `/private/tmp/cugraph-gnn-ruff/bin/ruff format --check` on changed files
- `/private/tmp/cugraph-gnn-ruff/bin/ruff check` on changed files
- `PYTHONPYCACHEPREFIX=/private/tmp/cugraph-pyg-pycache python3 -m compileall -q python/cugraph-pyg/cugraph_pyg`
- GPU integration test added but not run locally because the RAPIDS/PyTorch GPU dependencies are unavailable

Depends on #524 and rapidsai/cugraph#5636.

PR #524, opened 2026-08-21
Add fixed-window temporal sampling to cuGraph-PyG

alexbarghi-nv:
## Summary

- replace the homogeneous/heterogeneous, uniform/biased, and temporal/non-temporal sampling dispatch table with the unified `pylibcugraph.neighbor_sample` API added by rapidsai/cugraph#5625
- preserve the PyG `input_time` API while mapping it to the appropriate lower or upper temporal bound
- add `input_start_time` and `input_end_time` as mutually exclusive alternatives that activate fixed-window sampling
- propagate time windows through node batching, shuffling, and distributed call groups
- add homogeneous fixed-window coverage for uniform and biased sampling

## Testing

- `ruff format` and `ruff check` on changed files
- `python3 -m compileall -q python/cugraph-pyg/cugraph_pyg`
- GPU integration test added but not run locally because the RAPIDS/PyTorch GPU dependencies are unavailable in this environment

Depends on the cuGraph unified sampling and fixed-window support from rapidsai/cugraph#5603 and rapidsai/cugraph#5625.

PR #530, opened 2026-09-02
Add fixed-window and last temporal neighbor sampling to cuGraph-PyG

alexbarghi-nv:
## Summary

Consolidates and completes the two temporal sampling feature branches (#524 and #527) into a single PR, with additional validation, test coverage, and code cleanup applied on top.

### New features

**Fixed-window temporal sampling** (previously #524)
Adds `input_start_time` / `input_end_time` to `NeighborLoader`, allowing each seed node to carry its own `[start, end]` time window. The window is re-applied at every hop, so only edges whose timestamp falls within the seed's original window are eligible — this matches PyG's `temporal_strategy='uniform'` fixed-window behaviour.

**`temporal_strategy='last'`** (previously #527)
Adds `temporal_strategy='last'` to `NeighborLoader`. Among all eligible edges in the fixed time window, it deterministically keeps the fanout-K with the highest timestamps (most-recent neighbours), matching PyG's `temporal_strategy='last'`.

Requires `libcugraph` built from rapidsai/cugraph#5636.

### Changes on top of the original branches

- `fixed_window=True` now automatically enables `temporal=True` in `DistributedNeighborSampler` instead of requiring both flags.
- Fixed-window sampling is restricted to `temporal_comparison='monotonically_increasing'`; any other value raises `ValueError` at construction time in both `NeighborLoader` and `DistributedNeighborSampler`.
- Updated `NeighborLoader` docstring to document the fixed-window ordering constraint.
- `LinkNeighborLoader`: fixed `'suppported'` typo, upgraded the unsupported `temporal_strategy` warning to `ValueError`, and updated the docstring to direct users to `NeighborLoader` for `'last'` / fixed-window.
- `BaseDistributedSampler.sample_batches` abstract signature: renamed `time` → `seed_times`, added `seed_start_times` / `seed_end_times`, removed the dead `assume_equal_input_size` parameter.
- Added `test_neighbor_loader_temporal_last_unsupported_options` covering nine `ValueError` paths.
- Added three new `test_distributed_sampler` tests: invalid construction options for `'last'`, `fixed_window` auto-temporal, and `sample_batches` time-argument guards.

Closes #524, closes #527.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


python/cugraph-pyg/cugraph_pyg/sampler/distributed_sampler.py
```diff
@@ -817,13 +845,55 @@ def __init__(
...
+                    "The 'last' temporal strategy does not support biased sampling"
+                )
+
         # pylibcugraph requires temporal sampling to be disjoint.
         disjoint = disjoint or temporal
 
         self.__fanout = fanout
+        self.__fixed_window = fixed_window
+        self.__temporal_comparison = (
+            temporal_comparison or "monotonically_decreasing" if temporal else None
```
tingyu66, line 895: Should we change the default value from mono-decreasing to increasing as fixed window requires that? A constructor call with only (fixed_window = True) leads to "decreasing" currently.

alexbarghi-nv: I think this default was set to match the default behavior of PyG, but as long as that's enforced by `NeighborLoader` and `LinkNeighborLoader` we could change it.  I still wouldn't though - most people are going to expect monotonically decreasing to be the default, because it is the default in most GNN libraries.  It makes sense to have it fail with the default "monotonically decreasing" value so users are forced to be conscious of the requirement.

tl;dr I wouldn't change it.

tingyu66: That makes sense.

tingyu66 (approved):

